PR #1, opened 2023-07-03
Azure endpoint

trangdata:
- [x] add `deployment_name` and `model_name` for Azure endpoint
- [x] simple test


slobentanzer:
Thanks! Although this is not functional, I am merging to not have to handle multiple branches, as the separate class does not break any existing code.

trangdata:
After speaking with my admin, I think I got this to work now with the correct API version.

PR #14, opened 2023-07-31
Support to similarity search across all uploaded documents and map metadata to embeddings

fengsh27:
### Abstract
This submission is to support similarity search across all documents. To accomplish this, we introduced two vector collections,
1. metadata collection, save document meta data, its schema,
    Fields: ["id", "name", "author", "title", "format", "subject", "creator", "producer", "creationDate", "modDate", "source", "isDeleted"]
2. embedding collection, save documents embeddings, its schema,
   Fields: ["meta_id", "vector"]
Moreover, with "meta_id" in embeddings, we can obtain its document meta data.

Besides, in this submission, we also introduced DocumentEmbedder.save_document(), which encapsulated split_document() and store_embeddings() and would return document id.

### Exposed APIs
In **DocumentEmbedder**, we introduced these APIs:
- connect(host, port), connect to local/remote milvus server, and read all existed collections
- save_document(doc), save document to vector database
- similarity_search(query, k), similarity search by query text in all documents
- remove_document(doc_id), delete specified document
- get_all_documents(), retrieve all uploaded documents

### DIFF
--- a/biochatter/vectorstore_host.py
+++ b/biochatter/vectorstore_host.py
@@ -41,6 +41,12 @@ def align_embeddings(docs: List[Document], meta_id: int) -> List[Document]:
     return ret
 
 class VectorDatabaseHostMilvus:
+    """
+    VectorDatabaseHostMilvus is to manage vector databases in a host:
+    In it, we will manage embedding collection `_col_embeddings:langchain.vectorstores.Milvus` and 
+    metadata collection `_col_metadata:pymilvus.Collection`.
+
+    """
     def __init__(
         self,
         embedding_func: Optional[OpenAIEmbeddings] = None,
@@ -59,6 +65,13 @@ class VectorDatabaseHostMilvus:
         self._metadata_name = metadata_collection_name or DOCUMENT_METADATA_COLLECTION_NAME
     
     def connect(self, host: str, port: str) -> None:
+        """
+        Connects to a host and read two document collections (the default names are: DocumentMetadata 
+        and DocumentEmbeddings), if document collections don't exist, we will create the two collections.
+        Parameters:
+            host str: host ip address
+            port str: host port
+        """
         self._connect(host, port)
         self._init_host()
 
@@ -67,10 +80,18 @@ class VectorDatabaseHostMilvus:
         self.alias = self._create_connection_alias(host, port)
     
     def _init_host(self) -> None:
+        """
+        Initialize host. Will read/create document collection
+        """
         self._create_collections()
 
 
     def _create_connection_alias(self, host: str, port: str) -> str:
+        """
+        connect to host and create a connection alias for metadata collection
+        Returns:
+            str: connection alias
+        """
         alias = uuid.uuid4().hex
         try:
             connections.connect(host=host, port=port, alias=alias)
@@ -104,7 +125,10 @@ class VectorDatabaseHostMilvus:
         self._col_metadata.load()
         
 
-    def _create_metadata_collection_index(self):
+    def _create_metadata_collection_index(self) -> None:
+        """
+        create index for metadata collection
+        """
         if not isinstance(self._col_metadata, Collection) or \
             len(self._col_metadata.indexes) > 0:
             return
@@ -124,6 +148,13 @@ class VectorDatabaseHostMilvus:
             raise e
         
     def _create_metadata_collection_directly(self) -> None:
+        """
+        Create metadata collection.
+        All fields: "id", "name", "author", "title", "format", "subject", "creator", "producer", 
+        "creationDate", "modDate", "source", "embedding", "isDeleted"
+        As vector database requires one vector field, we will create a fake vector "embedding".
+        Field "isDeleted" is used to specify if the document is deleted.
+        """
         doc_id = FieldSchema(
             name="id",
             dtype=DataType.INT64,
@@ -202,6 +233,10 @@ class VectorDatabaseHostMilvus:
             raise e
             
     def _create_embeddings_collection_directly(self) -> None:
+        """
+        Create embedding collection.
+        All fields: "meta_id", "vector"
+        """
         try:
             self._col_embeddings = Milvus(
                 embedding_function=self._embedding_func,
@@ -213,7 +248,15 @@ class VectorDatabaseHostMilvus:
             raise e
         
 
-    def _insert_data(self, documents: List[Document]) -> Optional[str]:
+    def _insert_data(self, documents: List[Document]) -> str:
+        """
+        Insert documents to database.
+        Parameters:
+            documents List[Documents]: documents array, usually from DocumentReader.load_document
+                                       DocumentReader.document_from_pdf, DocumentReader.document_from_txt
+        Returns:
+            str: document id
+        """
         if (len(documents) == 0):
             return None
         metadata = [documents[0].metadata]
@@ -238,14 +281,29 @@ class VectorDatabaseHostMilvus:
         return meta_id
 
     def store_embeddings(self, documents: List[Document]) -> str:
+        """
+        Store documents to database.
+        Arguments;
+            documents List[Documents]: documents array, usually from DocumentReader.load_document
+                                       DocumentReader.document_from_pdf, DocumentReader.document_from_txt
+        Returns:
+            str: document id
+        """
         if len(documents) == 0:
             return
         return self._insert_data(documents)
-    def _build_embedding_search_expression(self, meta_query_res: List[Dict]) -> Optional[str]:
-        if len(meta_query_res) == 0:
+    def _build_embedding_search_expression(self, meta_ids: List[Dict]) -> Optional[str]:
+        """
+        Build search expression for embedding collection. The generated expression is like: "meta_id in [{id1}, {id2}, ...]
+        Parameters:
+            meta_ids: the array of metadata id in metadata collection
+        Returns:
+            str: search expression or None
+        """
+        if len(meta_ids) == 0:
             return None
         built_expr = '''meta_id in ['''
-        for item in meta_query_res:
+        for item in meta_ids:
             id = f'"{item["id"]}",'
             built_expr += id
         built_expr = built_expr[:-1]
@@ -253,6 +311,14 @@ class VectorDatabaseHostMilvus:
         return built_expr
     
     def _join_embedding_and_metadata_results(self, result_embedding: List[Document], result_meta: List[Dict]) -> List[Document]:
+        """
+        Join the search results of embedding collection and results of metadata.
+        Parameters:
+            result_embedding List[Document]: search result of embedding collection
+            result_meta List[Dict]: search result of metadata collection
+        Returns:
+            List[Document]: combined results like [{page_content: str, metadata: {...}}]
+        """
         def _find_metadata_by_id(metadata: List[Dict], id: str) -> Optional[Dict]:
             for d in metadata:
                 if str(d["id"]) == id:
@@ -269,9 +335,22 @@ class VectorDatabaseHostMilvus:
                 Document(page_content=res.page_content, metadata=found)
             )
         return joined_docs
+    
     def similarity_search(
         self, query: str, k: int=3
     ) -> List[Document]:
+        """
+        Similarity search according to query
+        This method will:
+        1). get all non-deleted meta_id and build to search expression for embedding collection
+        2). do similarity search in embedding collection
+        3). combine metadata and embeddings
+        Parameters:
+            query str: query string
+            k: the number of result
+        Returns:
+            List[Document]: search results
+        """
         result_metadata = self._col_metadata.query(
             expr="isDeleted == false"
         )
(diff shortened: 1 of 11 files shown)


slobentanzer:
Looking good! For the future, I have added `bump2version` to the repo, so we can use that to increment the version and automatically commit and tag (e.g. `bumpversion patch`).

slobentanzer:
@fengsh27 thanks for using bumpversion. I forgot to mention, since it autogenerates a git tag as well, you should also push the tag.

biochatter/vectorstore_host.py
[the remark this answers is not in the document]
slobentanzer: @fengsh27 here :)

test/test_vectorstore_host.py
```diff
@@ -42,52 +40,76 @@ def prepare_splitted_documents(
...
-        documents=splitted_docs,
-        embedding=OpenAIEmbeddings(),
-        connection_args={"host": _HOST, "port": _PORT},
-    )
-    utility.create_alias(db.collection_name, f"{docname}_{db.collection_name}")
-    collection_names.append(db.collection_name)
-
+    pass
 
 def teardown_module(module):
```
slobentanzer, line 45: @fengsh27 where/when is the teardown performed? what happens if only one test function is called and not the entire module? should we use a pytest fixture to ensure proper teardown after the tests are run?

slobentanzer:
I tried simply listing the connections after connecting through the `VectorDatabaseHostMilvus` class, but I am getting an error:

```
>>> utility.list_collections()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/slobentanzer/GitHub/biochatter/.venv/lib/python3.10/site-packages/pymilvus/orm/utility.py", line 501, in list_collections
    return _get_connection(using).list_collections(timeout=timeout)
  File "/Users/slobentanzer/GitHub/biochatter/.venv/lib/python3.10/site-packages/pymilvus/orm/utility.py", line 165, in _get_connection
    return connections._fetch_handler(alias)
  File "/Users/slobentanzer/GitHub/biochatter/.venv/lib/python3.10/site-packages/pymilvus/orm/connections.py", line 406, in _fetch_handler
    raise ConnectionNotExistException(message=ExceptionsMessage.ConnectFirst)
pymilvus.exceptions.ConnectionNotExistException: <ConnectionNotExistException: (code=1, message=should create connect first.)>
```

Not sure what goes wrong; the other functions related to getting the handler and connections seem to all be OK:

```
>>> dbHost.alias
'a01fd2ea0de3447eaad03e635e850208'
>>> connections.list_connections()
[('default', None), ('a01fd2ea0de3447eaad03e635e850208', <pymilvus.client.grpc_handler.GrpcHandler object at 0x1264aa860>)]
>>> utility._get_connection(dbHost.alias)
<pymilvus.client.grpc_handler.GrpcHandler object at 0x1264aa860>
```

Somehow the alias registration does not work when calling the `list_collections` function. Any ideas?

biochatter/vectorstore_host.py
```diff
@@ -1,196 +1,541 @@
...
+        )
+        self._col_metadata.load()
+
+    def _create_metadata_collection(self) -> None:
+        """
+        Create metadata collection.
+        All fields: "id", "name", "author", "title", "format", "subject", "creator", "producer",
+        "creationDate", "modDate", "source", "embedding", "isDeleted"
+        As vector database requires one vector field, we will create a fake vector "embedding".
+        Field "isDeleted" is used to specify if the document is deleted.
```
slobentanzer: @fengsh27 I am not 100% sure how the deletion process works. Do we only mark deleted embeddings / documents, meaning they aggregate over time, or do they actually disappear from the DB at some point? The latter would be preferable.

biochatter/vectorstore_host.py
```diff
@@ -1,196 +1,541 @@
...
     connections,
     Collection,
     utility,
+    DataType,
     FieldSchema,
+    CollectionSchema,
 )
 from langchain.vectorstores import Milvus
 from langchain.embeddings import OpenAIEmbeddings
 from langchain.schema import Document
```
slobentanzer, line 14: @fengsh27 is this module meant to completely replace the `vectorstore.py` module, i.e. the `DocumentEmbedder` and `DocumentReader` classes?

test/test_vectorstore_host.py
fengsh27: > @fengsh27 where/when is the teardown performed? what happens if only one test function is called and not the entire module? should we use a pytest fixture to ensure proper teardown after the tests are run?

setup_module/teardown_module is module level. Per my understanding, teardown will be called once all the tests in this module are executed (even if only one test in this module is executed). Of course, pytest fixture is more powerful.

fengsh27:
> I tried simply listing the collections in the DB after connecting through the `VectorDatabaseHostMilvus` class, but I am getting an error:
> 
> ```
> >>> utility.list_collections()
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "/Users/slobentanzer/GitHub/biochatter/.venv/lib/python3.10/site-packages/pymilvus/orm/utility.py", line 501, in list_collections
>     return _get_connection(using).list_collections(timeout=timeout)
>   File "/Users/slobentanzer/GitHub/biochatter/.venv/lib/python3.10/site-packages/pymilvus/orm/utility.py", line 165, in _get_connection
>     return connections._fetch_handler(alias)
>   File "/Users/slobentanzer/GitHub/biochatter/.venv/lib/python3.10/site-packages/pymilvus/orm/connections.py", line 406, in _fetch_handler
>     raise ConnectionNotExistException(message=ExceptionsMessage.ConnectFirst)
> pymilvus.exceptions.ConnectionNotExistException: <ConnectionNotExistException: (code=1, message=should create connect first.)>
> ```
> 
> Not sure what goes wrong; the other functions related to getting the handler and connections seem to all be OK:
> 
> ```
> >>> dbHost.alias
> 'a01fd2ea0de3447eaad03e635e850208'
> >>> connections.list_connections()
> [('default', None), ('a01fd2ea0de3447eaad03e635e850208', <pymilvus.client.grpc_handler.GrpcHandler object at 0x1264aa860>)]
> >>> utility._get_connection(dbHost.alias)
> <pymilvus.client.grpc_handler.GrpcHandler object at 0x1264aa860>
> >>> connections._fetch_handler(dbHost.alias)
> <pymilvus.client.grpc_handler.GrpcHandler object at 0x1264aa860>
> ```
> 
> Somehow the alias registration does not work when calling the `list_collections` function. Any ideas?

I think the issue is alias. utility.list_collection() uses "default" as default alias if "using" argument was not specified.

biochatter/vectorstore_host.py
```diff
@@ -1,196 +1,409 @@
...
+        self._connect(host, port)
+        self._init_host()
+
+    def _connect(self, host: str, port: str) -> None:
         self._connection_args = {"host": host, "port": port}
-        self._init_host(host, port)
+        self.alias = self._create_connection_alias(host, port)
+    
+    def _init_host(self) -> None:
+        """
```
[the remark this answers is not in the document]
fengsh27: > @fengsh27 can we explain here why we need to create an alias?

If here we don't specify an alias, the default alias "default" will be assigned to this connection. I prefer using our own alias.

biochatter/vectorstore_host.py
```diff
@@ -1,196 +1,409 @@
...
+        1). get all non-deleted meta_id and build to search expression for embedding collection
+        2). do similarity search in embedding collection
+        3). combine metadata and embeddings
+        Parameters:
+            query str: query string
+            k: the number of result
+        Returns:
+            List[Document]: search results
+        """
+        result_metadata = self._col_metadata.query(
```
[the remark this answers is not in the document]
fengsh27: > @fengsh27 is there a way to actually delete embeddings in the milvus system? if deleted entities are always just denoted as "isDeleted", i would imagine that the database will grow to unnecessarily large dimensions after a while. seems suboptimal for the 'long-running DB' use case.

Good question. I chose to use an "isDeleted" flag rather than hard deleting for the reason that it allows the possibility to "undo" a deletion by simply flipping the flag back to false. This can be handy for recovering accidentally deleted data. 

For the large dimensions issue, I think we can introduce a new field named "deleteDate." This would enable us to regularly remove items marked as "isDeleted," specifically those with a "deleteDate" occurring before a designated threshold. 

What's your opinion?

biochatter/vectorstore_host.py
fengsh27: No, the intention of this module is not to replace `DocumentEmbedder`. From my perspective, `VectorDatabaseHostMilvus` functions as milvus driver, while the `DocumentEmbedder` class serves as the primary interface exposed to ChatGSE.  
The `DocumentEmbedder` takes charge of text splitting and delegates database operations to `VectorDatabaseHostMilvus`.  In the future, we may need to introduce other vector database drivers like `VectorDatabaseHostChroma`. It will be `DocumentEmbedder` to manage various vector databases.

test/test_vectorstore_host.py
fengsh27: It seems I understood your concerns. Actually, my intention is to ensure the thorough cleanup of the test database once all the tests within this module have been executed rather than performing cleanup after each test. Therefore, the `teardown_module` is exactly what I want.

slobentanzer: So we can't run the tests in isolation, since they partly depend on each other? See also #16.

biochatter/vectorstore_host.py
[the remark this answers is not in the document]
slobentanzer: That sounds like a good idea, I like the undo option. Doing a periodic cleanup could be fine, or even a manual one. Something like `.delete_unused()`.

slobentanzer:
> I think the issue is alias. utility.list_collection() uses "default" as default alias if "using" argument was not specified.

@fengsh27 looking at the pymilvus API reference, I don't see a way to specify `using` in `list_collections()`; there is no parameter for that. It only has a `timeout`.

test/test_vectorstore_host.py
fengsh27: You're right, test_similarity_search does depends on test_store_embeddings. I'll fix it.

slobentanzer:
> > > I think the issue is alias. utility.list_collection() uses "default" as default alias if "using" argument was not specified.
> > 
> > 
> > @fengsh27 looking at the pymilvus API reference, I don't see a way to specify `using` in `list_collections()`; there is no parameter for that. It only has a `timeout`.
> 
> @slobentanzer the code is here: https://github.com/milvus-io/pymilvus/blob/9acbc7edf0c114554e79a00bfe1387e946eb9ef3/pymilvus/orm/utility.py#L529C87-L529C87

@fengsh27 seems the docstring was incomplete. i filed a PR adding the `using` parameter there. so if this works using our alias, I would propose to add management functionality to the `VectorDatabaseHost` class, at least for viewing the collections, some stats, and deleting the entries marked for deletion. What do you think?

fengsh27:
> > > > I think the issue is alias. utility.list_collection() uses "default" as default alias if "using" argument was not specified.
> > > 
> > > 
> > > @fengsh27 looking at the pymilvus API reference, I don't see a way to specify `using` in `list_collections()`; there is no parameter for that. It only has a `timeout`.
> > 
> > 
> > @slobentanzer the code is here: https://github.com/milvus-io/pymilvus/blob/9acbc7edf0c114554e79a00bfe1387e946eb9ef3/pymilvus/orm/utility.py#L529C87-L529C87
> 
> @fengsh27 seems the docstring was incomplete. i filed a PR adding the `using` parameter there. so if this works using our alias, I would propose to add management functionality to the `VectorDatabaseHost` class, at least for viewing the collections, some stats, and deleting the entries marked for deletion. What do you think?

@slobentanzer That makes sense. I'll try to add some management functionality.

biochatter/vectorstore_host.py
fengsh27: Indeed, the `test_document_summariser` comprehensively covers all facets of the `VectorDatabaseHost` class, encompassing crucial functionalities such as `store_embedding`, `similarity_search`, `get_all_documents`, and `remove_document`.

fengsh27:
@slobentanzer When I was trying to implement **delete_unused()** based on what we discussed before. I encounter an issue and its fix will be complicated. As Milvus does not support to update a record. the only way to flip "isDeleted" in meta collection is to delete the record completely, mark it as deleted, and re-insert it. 
The problem is that when we re-insert after deletion, the record gets a new ID from Milvus. But these record IDs are referenced by embedding collection. So those references then become invalid when the ID changes. 
An alternative solution is to introduce a new collection like **DeletedDocuments** to record the deleted documents. But this approach is not straightforward either.
Therefore, I go back to fully removal solution, which will actually delete embeddings and meta from database. This approach is much cleaner and robust.  In the future, when we decide to support "undo" operation, we can figure out how to fix this issue.

slobentanzer:
> Therefore, I go back to fully removal solution, which will actually delete embeddings and meta from database. This approach is much cleaner and robust. In the future, when we decide to support "undo" operation, we can figure out how to fix this issue.

@fengsh27 absolutely agree, good solution; it makes sense to defer the more complicated to later. We may use a different vector database vendor later, where this may be easier. Depending on that, we could make recommendations regarding "short-term" and "long-term" use cases.

slobentanzer (approved):

PR #22, opened 2023-08-14
Prompts

slobentanzer:
module and tests to generate knowledge graph queries by using the biocypher schema config in conjunction with LLM function parameterisation


slobentanzer (commented):
Basic prompt generation by the LLM

PR #26, opened 2023-11-03
create ci for pull request

fengsh27:
### Abstract
This submission is trying to create ci for pull requests on main branch. To pass unit tests on ci, this submission also did the following modifications:
1. Introduced milvus-docker-compose.yml to run milvus server for ci.
2. made some modification to test files.


fengsh27:
@slobentanzer , to make ci run successfully, we need add secrets OPENAI_API_KEY (https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions)

.bumpversion.cfg
slobentanzer, line 1: I would defer the bumping to after the merge. If we bump on the PR, the GitHub auto-generation of changes does not work.

pyproject.toml
```diff
@@ -21,7 +21,7 @@ redis = "^4.5.5"
 retry = "^0.9.2"
 transformers = "^4.30.2"
 streamlit = { version = "^1.23.1", optional = true }
-gTTS = { version = "^2.3.2", optional = true }
```
slobentanzer, line 24: Why not leave text-to-speech optional?

test/test_llm_connect.py
```diff
@@ -121,4 +121,5 @@ def test_azure():
         base=os.getenv("AZURE_TEST_OPENAI_API_BASE"),
     )
 
-    assert convo.set_api_key(os.getenv("AZURE_TEST_OPENAI_API_KEY"))
```
slobentanzer, line 124: Is this a workaround for the fact that we don't have that available on the GitHub runner? I would prefer to do this with a conditional skip or better with mocking. This way, the meaning of the test changes.

test/test_prompts.py
```diff
@@ -17,6 +17,7 @@ def test_biocypher_prompts(ps):
     ]
     assert list(ps.relationships.keys()) == [
         "Phosphorylation",
+        "GeneToPhenotypeAssociation",
```
slobentanzer, line 20: No, this was correct to not be included. It tests the ability to discard irrelevant associations according to the schema configuration.

pyproject.toml
fengsh27: When the tests run on pull request, it would fail due to unknown module gTTS

test/test_prompts.py
fengsh27: But, strangely, in my end, the test failed due to "GeneToPhenotypeAssociation" not included.

fengsh27:
> Thanks for the PR, @fengsh27!
> 
> I am currently planning the migration of all actual LLM access steps to a dedicated benchmarking setup, which would include the prompt engineering test, some of the podcast and LLM connection tests, basically everything that involves an actual response from the LLM.
> 
> Thus, it would probably make sense to modify the tests in CI to either only test the connectivity (setting the API key and getting the correct response from the API) or to mock an LLM response if need be. Then, we would not incur too many charges from OpenAI just for running the CI.
> 
> The benchmarking module, on the other hand, would run regularly to assess the actual performance of the LLMs in the tasks we give them, for which we need some benchmarking datasets.
> 
> Can we modify the CI tests to perform only those minimal connectivity tests and mock things such as responses and embeddings?

Sure, that makes sense. I'll start with vector database and vector database host to try to mock openai and milvus server (I think milvus server makes our unit test heavy). Is that OK?

.bumpversion.cfg
fengsh27: Got it. So, github will bump the version for us, right?

slobentanzer: Not exactly; however, bumping the version on main allows auto-generation of the version changes, which does not work if you do it on a PR already. It is some GitHub thing I think. So the procedure is work out the PR, merge, and then bump.

pyproject.toml
slobentanzer: Yeah, because I install with `-E podcast`, which installs the extra dependencies configured in the pyproject. However, if we mock things anyway, this could be one thing to mock. (It is not the expensive part of the test though.)

test/test_prompts.py
slobentanzer: Interesting; we should look into that in the benchmarking module. Not in the tests however, this is LLM behaviour and thus not reliable → should be mocked.

slobentanzer:
> I'll start with vector database and vector database host to try to mock openai and milvus server (I think milvus server makes our unit test heavy). Is that OK?

Yes, that's perfect. Thanks :)

test/test_prompts.py
slobentanzer: you were right, this is a different test than I originally thought. adding the association is correct.

test/test_llm_connect.py
fengsh27: Yes, this is a wrokaround, has been reverted back

pyproject.toml
fengsh27: Thanks. Currently, I added "install -E podcast" to ci.yml.

test/test_llm_connect.py
fengsh27: To make tests run successfully, besides OPENAI_API_KEY, we need to add the following env to github secrets:
AZURE_TEST_OPENAI_MODEL_NAME=
AZURE_TEST_OPENAI_DEPLOYMENT_NAME=
AZURE_TEST_OPENAI_API_VERSION=
AZURE_TEST_OPENAI_API_BASE=
AZURE_TEST_OPENAI_API_KEY=

fengsh27:
@slobentanzer , I think we can try to merge this pr first, so that, the CI can function properly in biochatter. And I'll continue to work on mocking job and raise another pr.

fengsh27:
BTW, it's true that GPT-4 is expensive, I think we can consider skipping GPT-4 tests.

slobentanzer (approved):
Thanks! Merging to test PR CI.

PR #49, opened 2023-11-28
Mock OpenAI and Milvus server for test_vectordatabase and test_vectordata_host

fengsh27:
### Abstract
As it is expensive to call OpenAI functions and communicate with milvus server in tests, this PR is trying to remove dependency on OpenAI and milvus server for the tests:`test_vectordatabase.py` and `test_vectordatabase_host.py`

After this PR, we will continue to remove dependency on OpenAI and milvus server in other tests.


test/test_vectorstore.py
```diff
@@ -64,8 +114,9 @@ def test_retrieval_augmented_generation_generic_api():
         ),
         embedding_collection_name="xinference_embedding_test",
         metadata_collection_name="xinference_metadata_test",
+        connection_args={"host": _HOST, "port": _PORT}
     )
-    rag_agent.connect(_HOST, _PORT)
```
fengsh27, line 68: @slobentanzer , it's strange that it will throw exception in XinferenceDocumentEmbedder constructor when I run tests today. 
```
# vectorstore.py
308        if self.model_name is None or self.model_name == "auto":
309            self.model_name = self.list_models_by_type("embedding")[0]
```
It works well yesterday. Then, I tried to get models from http://llm.biocypher.org with postman, Here is the returned values:
```
{

    "a823319a-88bd-11ee-8c78-0242acac0302": {
        "model_type": "LLM",
        "address": "0.0.0.0:46237",
        "accelerators": [
            "0"
        ],
        "model_name": "llama2-13b-chat-hf",
        "model_lang": [
            "en"
        ],
        "model_ability": [
            "embed",
            "generate",
            "chat"
        ],
        "model_description": null,
        "model_format": "pytorch",
        "model_size_in_billions": 13,
        "quantization": "none",
        "model_hub": "huggingface",
        "revision": null,
        "context_length": 4096
    },
    "23924b80-8c70-11ee-9f1b-a13cbc7efe14": {
        "model_type": "LLM",
        "address": "0.0.0.0:39351",
        "accelerators": [
            "1"
        ],
        "model_name": "llama-2",
        "model_lang": [
            "en"
        ],
        "model_ability": [
            "generate"
        ],
        "model_description": "Llama-2 is the second generation of Llama, open-source and trained on a larger amount of data.",
        "model_format": "pytorch",
        "model_size_in_billions": 13,
        "quantization": "none",
        "model_hub": "huggingface",
        "revision": "db6b8eb1feabb38985fdf785a89895959e944936",
        "context_length": 4096
    }
}
```
So, I'm wondering if we changed model_ability of "a823319a-88bd-11ee-8c78-0242acac0302" from "embedding" to "embed", which makes `self.list_models_by_type("embedding")` return empty array.

slobentanzer: @fengsh27 good question; the models returned by llm.biocypher.org can change depending on what @AndiMajore and colleagues run on their node (which the URL redirects to). I am not sure whether there is a consensus to use "embed" or "embedding" in Xinference, or even whether Xinference is the framework that defines this. @AndiMajore do you have insights there?

AndiMajore: Hey @fengsh27, @slobentanzer  is right, we from time to time restart and use different models, and with every restart a new model_uid is given. Just yesterday we changed our setup to now run two LLMs and one text-embedding model (now on CPU), but that was tested and worked, so should not be the cause of this issue. While doing that I also realized that there were olny two LLMs running and no embedding-model, which looking at your [http://llm.biocypher.org/v1/models](http://llm.biocypher.org/v1/models) output shows exactly that. This should be fixed by now, but it seems that someone shut down the text-embedding model to start a llama-2 instance instead, which was not intended.
Because for now we do not have resources to support LLM-based embedding well, we try to rely on way faster and more lightweight models for that, so it is very convenient that Xinference is providing not only LLMs but also text-embedding models.

This is also why we have this function that lists all available models by type and name instead of using a hard-coded UID.
Regarding the difference between "embed" and "embedding": a text-embedding model will be labeled with "embedding" whereas a LLM that supports embedding will be labeled as "embed".

Here a screenshot of the currently running models:

image

fengsh27: @AndiMajore Thanks for your explanation. Now I understand the difference between "embed" and "embedding".

fengsh27: @slobentanzer  I think it make sense to mock them in tests.

slobentanzer:
BTW, this means that the test battery needs to be run with `pytest test`, specifying the directory, instead of just `pytest`. I did already implement that in the CI yaml, here: https://github.com/biocypher/biochatter/blob/e7bc13ab087b3401867bdcf49a912ac8acaee96a/.github/workflows/ci.yaml#L60

slobentanzer (approved):
Really great work and a big step forward in the CI of BioChatter! Thanks :)

PR #65, opened 2023-12-13
Test_multi_word

slobentanzer:
Merge multi-word benchmarks into benchmark branch


slobentanzer:
@YY-SONG0718 I created a PR with the benchmark branch as target, because that gives us a good venue to discuss individual aspects of the changes. I will start a review and comment at specific locations.

benchmark/test_biocypher_query_generation.py
```diff
@@ -11,19 +13,31 @@
     None,
 )
 
-MODEL_NAMES = [
+OPENAI_MODEL_NAMES = [
     "gpt-3.5-turbo",
     "gpt-4",
 ]
 
+XINFERENCE_MODEL_NAMES = [
```
slobentanzer, line 21: I added these as a placeholder to figure out how we will manage starting, stopping, and calling different models from within the pytest process. Selecting these does not currently work, so running `pytest benchmark` will give errors for these parts of the matrix.

benchmark/test_biocypher_query_generation.py
```diff
@@ -11,19 +13,31 @@
...
-    return BioCypherPromptEngine(
-        schema_config_or_info_path="test/test_schema_info.yaml",
-        model_name=model_name,
-    )
+    if model_name in OPENAI_MODEL_NAMES:
+        return BioCypherPromptEngine(
+            schema_config_or_info_path="test/test_schema_info.yaml",
+            model_name=model_name,
+        )
+    elif model_name in XINFERENCE_MODEL_NAMES:
```
slobentanzer, line 37: Here would be a potential place to deal with the Xinference model integration

benchmark/test_biocypher_query_generation.py
```diff
@@ -140,15 +150,300 @@ def test_end_to_end_query_generation(prompt_engine):
...
+            f"{prompt_engine.model_name},end-to-end_single,{calculate_test_score(score)}\n"
+        )
+
+
+######
+# test multiple word entities
+######
+
+
+# entity selection doesn't have any issue
```
slobentanzer, line 168: Fine for now, but I would not expect this to hold for all models. In the context of the benchmark across models, prompts, etc, the scores will be the measure of whether it works or not.

benchmark/test_biocypher_query_generation.py
```diff
@@ -140,15 +150,300 @@ def test_end_to_end_query_generation(prompt_engine):
...
+    # make sure conditioned
+    score.append("WHERE" in query or "{cell_type_name:" in query)
+
+    with open(FILE_PATH, "a") as f:
+        f.write(
+            f"{prompt_engine.model_name},cypher query_single,{calculate_test_score(score),query}\n"
+        )
+
+
+def test_end_to_end_query_generation_multi_word(prompt_engine):
```
slobentanzer, line 300: not sure we should do end-to-end tests for all scenarios, maybe the simple one is enough. in theory, this should be covered by the step-wise tests, but maybe keeping one of them is fine. many of those could needlessly increase cost/runtime.

benchmark/test_biocypher_query_generation.py
```diff
@@ -140,15 +150,300 @@ def test_end_to_end_query_generation(prompt_engine):
...
+    entity_mapping = map_entities_to_labels(used_entities)
+
+    # get all the entity and respective properties used in the cypher query
+    used_entity_property = join_dictionaries(entity_mapping, property_mapping)
+
+    return entity_mapping, property_mapping, used_entity_property
+
+
+def test_property_exists(prompt_engine):
+    query = prompt_engine.generate_query(
```
slobentanzer, line 409: I think this should not be the end-to-end process, but rather just the property selection step. This way is less controlled and also costs more.

slobentanzer (commented):
Definitely a good start :)

benchmark/test_biocypher_query_generation.py
YY-SONG0718: Sounds good!

benchmark/test_biocypher_query_generation.py
YY-SONG0718: Removed this comment after clean up

benchmark/test_biocypher_query_generation.py
YY-SONG0718: Only kept single word end-to-end test.

benchmark/test_biocypher_query_generation.py
YY-SONG0718: Changed this test to start by selecting properties. The model now tends to invent a new property in the query if it does not pick any properties that it thinks are relevant.

YY-SONG0718:
@slobentanzer I made the changes as discussed. Could you please have a look when you are available?

slobentanzer:
Thanks! :)

Work continues on the benchmark branch.

PR #71, opened 2023-12-14
Query interaction

emmaver:
First basis for query interaction using a `BioCypherQueryHandler` object. Allows to ask for an explanation, or request an update of a generated query.

Multiple todo's and improvements are left, but basic functionality is there.


slobentanzer (approved):
Great job, thanks :)

PR #73, opened 2023-12-14
parametrize benchmark tests based on test data in csv; enable switchi…

nilskre:
…ng between public test data in this repo and private test data; decryption process of the private test data


benchmark/conftest.py
```diff
@@ -31,6 +34,15 @@ def delete_csv_files():
             f.write("")
 
 
+def pytest_generate_tests(metafunc):
```
slobentanzer, line 37: do we want (need) to run this every time the fixture is requested, or would it be possible to define this on session or module scope?

benchmark/load_dataset.py
```diff
@@ -0,0 +1,101 @@
...
+from ast import literal_eval
+from base64 import b64decode
+
+import pandas as pd
+import rsa
+import yaml
+from cryptography.fernet import Fernet
+
+
+def get_all_files(directory: str) -> list[str]:
```
slobentanzer, line 13: docstrings

slobentanzer (commented):
Nice

benchmark/conftest.py
nilskre: After having a look at the [pytest docs](https://docs.pytest.org/en/latest/how-to/parametrize.html#pytest-generate-tests) and [this article](https://pytest-with-eric.com/introduction/pytest-generate-tests/):

This is expected behavior and pytest is designed to behave like this. The expected behavior is that pytest calls the `pytest_generate_tests` once for every test in the module in the initial test collection. If the test contains a fixture from the `pytest_generate_tests` function, the test is parameterized. Thus the number of test cases = number how often `pytest_generate_tests` is called to check if there is something to parametrize.

slobentanzer: But if the fixture content does not change, this could be just run once? I thought that's what fixture scopes are for...

slobentanzer: In other words, if you separate dataset loading (only once, since the dataset will not change) and function parameterisation (for each test), would that not make it more efficient?

nilskre: Yes, I see the point. I am trying to figure out how to configure this. Then the dataset itself would be a fixture and is loaded once and passed to the `pytest_generate_tests` (in theory - let's see how to implement this :)

nilskre: Fixed it.

benchmark/test_biocypher_query_generation.py
```diff
@@ -13,142 +19,284 @@
...
+        model_name (str): The model name, e.g. "gpt-3.5-turbo"
+        result_files (dict[str, pd.DataFrame]): The result files
+
+    Returns:
+
+        bool: True if the benchmark task and subtask for the model_name has
+            already been run, False otherwise
+    """
+    task_results = result_files[f"benchmark/results/{task}.csv"]
+    task_results_subset = (task_results["model"] == model_name) & (
```
slobentanzer, line 60: For me this fails because the key "model" is not in the DF.

benchmark/test_biocypher_query_generation.py
```diff
@@ -13,142 +19,284 @@
...
+    subtask = inspect.currentframe().f_code.co_name
+
+    kg_schema_file_name = test_data_biocypher_query_generation[0]
+    prompt = test_data_biocypher_query_generation[1]
+    expected_entities = test_data_biocypher_query_generation[2]
+    prompt_engine = prompt_engine(
+        kg_schema_path=f"./benchmark/data/biocypher_query_generation/{kg_schema_file_name}"
     )
+
+    if benchmark_already_executed(
```
slobentanzer, line 83: We may need more expressive subtask names. The current implementation generates two tests for the entity selection, one for each of the two rows in the DF. However, both are called the same, `test_entity_selection`. This indicates that 1) the check here does not work correctly (otherwise it would skip the second run with the same subtask name), and 2) we need some way to generate a unique subtask name for each row.

benchmark/test_biocypher_query_generation.py
```diff
@@ -13,142 +19,284 @@
...
+            expected_relationship_label_key
+        ]:
+            score.append(
+                expected_relationship_label_value
+                in prompt_engine.selected_relationship_labels[
+                    expected_relationship_label_key
+                ]
+            )
+
+    # TODO: make it more generic to be able to compare arbitrarily nested structures
```
slobentanzer, line 151: There are some improvements to the benchmarking of queries using regex in the PR I just merged into the `benchmark` branch by Yuyao; we should merge these functions here and use them to genericise the score generation. https://github.com/biocypher/biochatter/blob/f970a48ebd967539f19862b78af2a496e1af0356/benchmark/test_biocypher_query_generation.py#L386C8-L386C8

slobentanzer (commented):
The abstraction seems like it's on the right path, although I am not sure whether some parts are maybe too much engineering, just shifting the complexity (e.g. from the test to the CSV). We need to think about the correct level of abstraction for the task, for instance whether we need a completely separate CSV layout for each test module, and how we can genericise the scoring (expected vs found). I think we are close to a functional solution though. :)

benchmark/test_biocypher_query_generation.py
nilskre: Now fixed

benchmark/test_biocypher_query_generation.py
nilskre: `test_end_to_end_query_generation_0` for the first row, `test_end_to_end_query_generation_1` for the second row, and so on. So now the check is working correctly and checking on the subtask test case level (row) and not only on the subtask level (file).

slobentanzer:
Good work; merging :)

PR #87, opened 2024-01-08
Benchmark

nilskre:


slobentanzer:
@nilskre we have two PRs for the same two branches now, this and #73...

What is the plan, what is the merging order? Wouldn't either result in the same end state? Why do we need both? I originally though that `benchmark` should be the target branch that then is merged into `main` once stable.

PR #90, opened 2024-01-12
make rag to support embedded documents workspace

fengsh27:
### Abstract
This submission is to make rag support embedding documents workspace. Currently, all embedded documents operations such as obtaining all documents, performing similarity search and removing documents, are preformed across all embedded documents in vector store. With this submission, an embedded documents workspace (essentially a list of embedded document IDs) will be passed to rag, which is used to define the scope for all operations. If no workspace (i.e., `None`) passed in, all operations will be considered to be performed across all documents as before.

Besides introducing embedded documents workspace, this submission also moved xinference importing back to XinferenceDocumentEmbedder constructor.  Previously, the xinference importing was moved to the beginning of the file to facilitate unit tests. However, this unexpectedly increased the size of  biochatter-server docker image, even if xinference was not yet needed. By moving it back, it optimizes biochatter-server docker size and ensures that xinference is only imported when necessary.


slobentanzer (approved):
Looks good and makes sense, thanks! Will bump version to `0.3.11` after the merge.

PR #91, opened 2024-01-15
Reduce dependency on xinference

fengsh27:
### Abstract
In previous submission, to optimize dependency on xinference, I made changes in `vectorstore.py`, But I forgot to move xinference importing back to constructor of `XinferenceConversation` in `test_llm.py`, which result in biochatter-server's dependency on xinference. This submission is to fix this issue.


slobentanzer (approved):
Great, thanks! Will bump version to `0.3.12`

PR #92, opened 2024-01-18
Introduce AzureOpenAIEmbeddings for Azure OpenAI server

fengsh27:
### Abstract
After upgrading openai to >= 1.0.0, OpenAIEmbeddings can't be used with Azure OpenAI settings any more. This submission was to fix the issue by introducing AzureOpenAIEmbeddings.


slobentanzer (approved):
Thanks!

PR #97, opened 2024-01-23
Rag_agent

slobentanzer:
This PR introduces a new high-level API for doing RAG in a unified manner with either a KG (BioCypher) or a vector database (via pre-existing Milvus class). Basic usage is:

- instantiate `RagAgent`, giving it connection args and KG or vector DB parameters

- use `generate_responses` method to return, with an input of the user question and a number k of desired results, a list of tuples containing the RAG text as first elements (str) and the metadata as second elements (dict)

The dependency tree can be seen in the newly added drawio figure (docs folder). Of note, the VectorDatabaseHostMilvus was renamed to VectorDatabaseAgentMilvus, and the vectorstore_host module to vectorstore_agent, to correspond with the newly introduced modules and classes.


slobentanzer:
@fengsh27 please review with regard to whether it makes sense to have this class as the basis of the REST API from biochatter-server when performing RAG for either vector DB or KG. This would replace the current logic we use for the RAG that is already implemented. Is a list of tuples [(Rag str, metadata dict)] OK for streaming between biochatter and biochatter-server?

fengsh27:
Sure, I'll have a look ASAP

fengsh27:
@slobentanzer I have some questions regarding `RagAgent`,
1. According the drawio figure, I assume that `DocumentEmbedder` will be replaced by `RagAgent`, right?
2. Will `RagAgent` be integrated to llm_connect (GptConversation)?

slobentanzer:
> According the drawio figure, I assume that DocumentEmbedder will be replaced by RagAgent, correct? If not, it appears there may be some overlap in functionality between DocumentEmbedder and RagAgent.

Essentially yes; the `RagAgent` should be in charge of all RAG applications from anywhere else.

The RAG agent is supposed to be a higher-level access to the RAG functionalities (also with KG queries). It will, in some use cases, use the same function in the former `VectorDatabaseHostMilvus` (now "Agent") to do semantic search. I think the semantic search from the DocumentEmbedder was also just a call to the original method in the `VectorDatabaseHostMilvus`, correct?

The `RagAgent` is now a level above the vector database host, and simultaneously access the vector semantic search and the KG queries, with largely the same API. Using one API for both types of RAG is the main motivation for creating this parent module. My question was whether this makes sense from an architectural perspective, in the sense that querying it from `biochatter-server` may be easier if we only need one API for any RAG.

> Will RagAgent be integrated to llm_connect (GptConversation) or directly called by biochatter-server?

I think both could be possible, depending on where it is better for the use case. It could also be both, e.g. if a RAG process should be performed only in `biochatter` for use in a different frontend or in pure Python even. If you think that orchestrating it in `biochatter` vs `biochatter-server` would be approximately the same effort, it is probably more useful to do it in `biochatter` directly, as this can be used in other cases as well. If it is much easier to do it in `biochatter-server`, we could do that first.

Thoughts?

fengsh27:
@slobentanzer This looks good to me. In biochatter-server, we currently interact directly with `DocumentEmbedder` when performing operations on RAG, like connecting to vectordb, uploading doc, removing doc and listing all docs. On the other hand, during chat conversations, we synchronize RAG arguments (connection args, doc ids, etc.) with `DocumentEmbedder` in GptConversation. Here is a diagram:

image

fengsh27:
I think we can keep this workflow with RagAgent.

slobentanzer:
Yes, I think so too. In the longer run, we should generalise to open source models (exchange the GptConversation for a more generic one, and maybe inject the RagAgent instead of calling it from inside the conversation), but for now the first step is replacing the DocumentEmbedder in the GptConversation with the RagAgent. I can do that next.

fengsh27 (approved):
Looks good

fengsh27:
@slobentanzer shoud I work on a branch derived from branch `benchmark`?

slobentanzer:
Yes, that's probably the best option. I want to merge benchmark into main as soon as possible anyways; benchmark has the current changes to module structure.

fengsh27:
Got it.

PR #113, opened 2024-02-02
Inject conversation factory to RagAgent

fengsh27:
### Abstract
This submission was to inject conversation factory to RagAgent. For KG RagAgent, it uses conversation factory to create `Conversation` to request to generate Neo4j query.


slobentanzer (approved):
Great work, thank you!

PR #118, opened 2024-02-06
fix exception in kg query

fengsh27:
### Abstract
When kg query got no results (question was not related to kg), it would throw exception, which will break the chat. To fix this issue, we need to catch this exception and leave context empty.


biochatter/llm_connect.py
```diff
@@ -257,18 +260,26 @@ def _inject_context(self, text: str):
                 for agent in self.rag_agents:
                     if not agent.use_prompt:
                         continue
-                    statements = statements + [
-                        doc[0] for doc in agent.generate_responses(text)
-                    ]
+                    try:
+                        docs = agent.generate_responses(text)
+                        statements = statements + [
+                            doc[0] for doc in agent.generate_responses(text)
```
slobentanzer, line 266: shouldn't this be `for doc in docs`, as below?

fengsh27: Good catch.

biochatter/llm_connect.py
```diff
@@ -257,18 +260,26 @@ def _inject_context(self, text: str):
...
+                        ]
+                    except Exception as e:
```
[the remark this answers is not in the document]
fengsh27: make sense

slobentanzer (approved):
Thanks!

PR #124, opened 2024-02-13
In chatting, return injected contexts

fengsh27:
### Abstract
This submission was to implement returning injected context in chat


slobentanzer (approved):
Good work! :)

PR #157, opened 2024-05-24
[PDSM] Medical Testcases for benchmarking

ytehran:
Please note, that this PR is still in work

## What does this PR do?

- Creation of test cases split into medical fields and languages (German & English)
- Analyzing those results and classifying their failure groups

## Notes
For the analysis:
Adaption of benchmark_utils for the failure_groups and added new methods
To test for synonyms, “nltk” is used, which must
1. be installed e.g. pip install nltk
2. Downloaded by commenting line 26 in our test.py


.gitignore
```diff
@@ -4,6 +4,7 @@ dist/
 __pycache__/
 .venv
 .pytest_cache
+venv/.env
```
slobentanzer, line 7: not sure I understand why we have this. why not use `.venv`?

benchmark/conftest.py
```diff
@@ -1,29 +1,28 @@
 import os
 
 import requests
+from dotenv import load_dotenv
```
slobentanzer, line 4: Why do we need `dotenv`? The implementation seems a bit hacky ;)

benchmark/conftest.py
```diff
@@ -233,6 +232,9 @@ def conversation(request, model_name):
             prompts={},
             correct=False,
         )
+        # delete first dots if venv is in project env
```
slobentanzer, line 235: As mentioned, I would not do this. Rather use `.venv` and configure your setup to use this env reliably. Env setup is on the user, should not be mandated by the code.

openAIKey.py
```diff
@@ -0,0 +1,9 @@
+import os
```
slobentanzer, line 1: Again, we should not need this. I would suggest removing and finding a local solution to point the interpreter to the right env.

test/no_podcast.py
slobentanzer, line 1: Why did you rename this? Should be reverted.

benchmark/test_correctness_medical_usecase.py
```diff
@@ -0,0 +1,108 @@
...
+    skip_if_already_run,
+    get_result_file_path,
+    get_wrong_result_file_path,
+    write_results_to_file,
+    write_wrong_results_to_file,
+    categorize_failures
+)
+
+
+def test_correctness_of_answers(
```
slobentanzer, line 19: A docstring explaining the test would be nice. :)

benchmark/test_rag_interpretation.py
```diff
@@ -5,7 +5,7 @@
 from biochatter._misc import ensure_iterable
 from .conftest import calculate_test_score
 from .benchmark_utils import (
-    skip_if_already_run,
```
slobentanzer, line 8: Try not to interfere with existing functionalities in your PRs. I reverted it here.

slobentanzer (changes requested):
Thanks @ytehran and others, the dataset is pretty nice. I have some general comments about implementation that you could address/discuss; will look into the actual test algorithm next.

benchmark/benchmark_utils.py
```diff
@@ -96,6 +101,47 @@ def return_or_create_result_file(
     return results
 
 
+def return_or_create_wrong_result_file(task: str):
```
slobentanzer, line 104: This is not an intuitive name for the function. I had trouble understanding what it is for just from reading the code (what is a "wrong result file"). I would suggest naming the entire process something like "failure mode identification" such that it is intuitively clear what is happening. What you are doing is saving responses in case of a failure.

slobentanzer: I changed the naming in [c137e4c](https://github.com/biocypher/biochatter/pull/157/commits/c137e4c5508f4d25e687d70d200cc78510e6e283)

.gitignore
ytehran: Happened automatically, we removed it

benchmark/conftest.py
ytehran: Might be a bit of a work around, but this was the only way we could implement the key loading. We have some questions, later per mail

openAIKey.py
ytehran: Will reply to this in the following mail

benchmark/test_correctness_medical_usecase.py
```diff
@@ -0,0 +1,110 @@
...
+    model_name,
+    test_data_correctness,
+    conversation,
+    multiple_testing,
+):
+    # Downloads the naturale language synonym toolkit, just need to be done once per device
+    # nltk.download()
```
slobentanzer, line 26: This is not a good solution. Often, we execute runs on remote hardware or GitHub runners, where we can't "manually uncomment and install something."

I suggest you find a way to programmatically do this. IIRC, NLTK only downloads if it's not in cache. If not, I would probably resort to a Pytest fixture that checks whether it is there, and only downloads if it is not.

ytehran:
Wir haben jetzt einen Teil der SW Engineering umgesetzt und die Fragen anpasst. Wir wären ready um den großen Durchlauf zu machen

benchmark/conftest.py
slobentanzer: @ytehran could you or another team member address this? prevents me from merging.

benchmark/conftest.py
```diff
@@ -387,6 +386,9 @@ def evaluation_conversation():
         prompts={},
         correct=False,
     )
+    # delete first dots if venv is in project env
+    cus_path = os.getcwd() + "../../venv/bin/.env"
```
slobentanzer, line 390: @ytehran please someone remove this so we can merge this PR.

openAIKey.py
slobentanzer: @ytehran please address this (remove) in order to be able to merge into `main` :)

ytehran:
@slobentanzer We changed our code in order to get the PR ready. Even some updates and improvements for our stats. If something is missing, please let me know

PR #167, opened 2024-06-21
Multimodal model support

slobentanzer:
works fine with OpenAI
open source technically works, but the model (currently qwen-vl-chat) returns nonsense answers


drAbreu:
Hi Sebastian,

I see the issue closed, but I was taking a look to it. 

There are some points that I thought it could be worth to bring up to discussion. Mainly to avoid possible issues in the model upload.

There is only one thing I would say needs to be addressed, which is the format of the image files.

You have tried with `.jpg` and `.png` in the code (unless I missed other test case). However, in the OpenAI docs says that:

[What type of files can I upload?](https://platform.openai.com/docs/guides/vision/what-type-of-files-can-i-upload)
We currently support PNG (.png), JPEG (.jpeg and .jpg), WEBP (.webp), and non-animated GIF (.gif).

For this reason, I would suggest to add, on a series of functions that might be `utils` or a series of functions connected to `llm_connect.encode_image`, there should be a conversion from other formats, such as `pdf`, `eps` or `tif`, to either `png` or `jpeg`.  This implementation is something I have done on a very recent project, so it is not an issue at all to add it into the code. 

Another possible topic to consider, although not highly important for OpenAI, is the image size. It might be of interest to provide a `resize_image` function to re-scale images. This might be useful for open source models that are typically limited to smaller image sizes. That could make us sure to know what the input into the model is. This is also an easy implementation. 

I could get both of them done in a few hours if you think they make any sense.

For the rest, the implementation you have is very clean and straight forward, so nothing else to say aside of the two topics above. The only extra thing would of course be to add 2 or 3 tests to test the usage of image formats other than those supported by OpenAI.

slobentanzer:
Hi Jorge,
all good recommendations, thanks. The PR/issue are closed because the PR with basic functionality is merged. I agree that your suggestions are good; the way we handle this is to create new issues for them and add them via separate PR. Feel free to create any number of issues to account for your ideas and tag me in order to organise them in the project structure. PRs are accepted at any time, even without previous issue (if they are of this small scale).

PR #168, opened 2024-06-21
Rag api agent

Tatan47:
This pull request introduces a new mode, ‘API’, for the RAG agent, enabling it to connect with the NCBI BLAST database.

New API Agent:
	•	Translates natural language queries into NCBI BLAST API calls.
	•	Retrieves and saves BLAST responses.
	•	Extracts answers and returns them in natural language.
RAG Agent Update:
	•	Introduces a new mode, API, integrating the new API agent.
Live Tests:
	•	Comprehensive end-to-end tests for the new components.
Mock Tests:
	•	Mock API calls for both the LLM and NCBI BLAST

Areas Needing Assistance:
	•	Mock test
	•	Final integration of the API agent into the Conversation class


biochatter/api_agent.py
```diff
@@ -0,0 +1,441 @@
...
+
+    Attributes:
+        llm (object): The language model to be used.
+        blast_result_path (str): The path to save BLAST results.
+        blast_prompt_path (str): The path to the BLAST prompt file.
+        builder (BlastQueryBuilder): An instance to build BLAST queries.
+        fetcher (BlastFetcher): An instance to fetch BLAST results.
+    """    
+    def __init__(self, llm):
+        self.llm = llm
```
fengsh27, line 356: In constructors of `RagAgent` and `DatabaseAgent`, we will pass `conversation_factory()` function to create conversation. This is because, in some cases, BioChatter may not have knowledge to create LLM (ChatGPT or Azure ChatGPT).
I think here we can follow the approach of RAGAgent and DatabaseAgent to use conversation_factory() to create conversation, then utilize conversation.chat as llm

slobentanzer: Absolutely, good observation. Do you have time to suggest the refactor?

fengsh27: Yes, I recommend, in APIAgent's constructor, it would be better to pass in conversation_function instead of llm.

biochatter/api_agent.py
```diff
@@ -0,0 +1,441 @@
...
+        self.fetcher = BlastFetcher()
+        self.final_answer = None
+        self.error = None
+    
+        os.makedirs(self.blast_result_path, exist_ok=True)
+
+    def generate_blast_query(self, question: str) -> Optional[BlastQuery]:
+        try:
+            return self.builder.generate_blast_query(
+                question, self.blast_prompt_path, self.llm
```
fengsh27, line 369: And here, we can create llm like this:
```
conversation = self.conversation_factory()
return self.builder.generate_blast_query(question, self.blast_prompt_path, conversation.chat)
```

biochatter/rag_agent.py
```diff
@@ -92,9 +99,14 @@ def __init__(
             self.agent.connect()
 
             self.query_func = self.agent.similarity_search
+
+        elif self.mode == RagAgentModeEnum.API:
+            from .api_agent import APIAgent
+            llm = ChatOpenAI(model_name='gpt-4', temperature=0, openai_api_key=os.getenv("OPENAI_API_KEY"))
```
fengsh27, line 105: Here, instead of creating `llm` here, we'd better to pass in `conversation_factory` to `APIAgent`.

biochatter/api_agent.py
```diff
@@ -0,0 +1,441 @@
...
+    Attributes:
+        llm (object): The language model to be used.
+        blast_result_path (str): The path to save BLAST results.
+        blast_prompt_path (str): The path to the BLAST prompt file.
+        builder (BlastQueryBuilder): An instance to build BLAST queries.
+        fetcher (BlastFetcher): An instance to fetch BLAST results.
+    """    
+    def __init__(self, llm):
+        self.llm = llm
+        self.blast_result_path = ".blast"
```
slobentanzer, line 357: should be handled by the query builder

biochatter/api_agent.py
```diff
@@ -0,0 +1,441 @@
...
+        self.blast_prompt_path = "docs/api_agent/BLAST_tool/persistent_files/api_documentation/BLAST.txt"
```
slobentanzer, line 358: should be handled by the query builder, and ideally not be a path at all (but a string)

biochatter/api_agent.py
```diff
@@ -0,0 +1,441 @@
...
+        self.builder = BlastQueryBuilder()
```
slobentanzer, line 359: if the API agent is supposed to decide which tool to use, these would have to be generic

biochatter/api_agent.py
```diff
@@ -0,0 +1,441 @@
...
+                If you are asked what organism a specific sequence belongs to, check the 'Hit_def' fields. If you find a synthetic construct or predicted entry, move to the next one and look for an organism name.\n\
+                Try to use the hits with the best identity score to answer the question. If it is not possible, move to the next one.\n\
+                Be clear, and if organism names are present in ANY of the results, please include them in the answer. Do not make up information and mention how relevant the found information is based on the identity scores.\n\
+                Use the same reasoning for any potential BLAST results. If you find information that is manually curated, please use it and state it. You may also state other results, but always include the context.\n\
+                Based on the information given here:\n\
+                {context}
+                """
+        output_parser = StrOutputParser()
+        chain = prompt | llm | output_parser
+        BLAST_answer = chain.invoke(
```
slobentanzer, line 337: using the langchain idiomatic style is constraining, in this case probably unnecessarily. would refactor to be a simple query built from the parts

biochatter/api_agent.py
```diff
@@ -0,0 +1,441 @@
...
+                    lines.append(line.strip())
+                # to test:
+                # more efficient with \n or without?
+                return "\n".join(lines)
+        except FileNotFoundError:
+            return "The file was not found."
+        except Exception as e:
+            return f"An error occurred: {e}"
+
+    def answer_extraction(self, question: str, file_path: str, n: int) -> str:
```
slobentanzer, line 302: would rather put this into a third class, the Interpreter (in addition to the generic QueryBuilder and Fetcher classes)

slobentanzer (changes requested):

biochatter/api_agent.py
slobentanzer: implemented

biochatter/api_agent.py
slobentanzer: implemented

biochatter/rag_agent.py
slobentanzer: implemented

slobentanzer (commented):
We might have to account for two types of API:
- BLAST as the initial example has a two-step API, submission of the request and retrieval of the results
- many APIs have a one-step process, where the return of the API request is the result

The Fetcher base class (in `api_agent/abc.py`) may need to account for this. Maybe by implementing a wrapper that can unify both cases.

biochatter/api_agent/api_agent.py
```diff
@@ -0,0 +1,154 @@
...
+            return None
+
+    def summarise_results(self, question: str, file_name: str) -> Optional[str]:
+        """
+        Summarise the retrieved results to extract the answer to the question.
+        """
+        try:
+            file_path = os.path.join(self.result_path, file_name)
+            return self.result_interpreter.summarise_results(
+                question, file_path, 100
```
fengsh27, line 104: missing parameter conversation_factory?

biochatter/api_agent/blast.py
```diff
@@ -0,0 +1,352 @@
...
+        # Check the status of the BLAST job
+        for attempt in range(max_attempts):
+            status_response = requests.get(base_url, params=check_status_params)
+            status_response.raise_for_status()
+            status_text = status_response.text
+            print("evaluating status")
+            if "Status=WAITING" in status_text:
+                print(f"{question_uuid} results not ready, waiting...")
+                time.sleep(15)
+            elif "Status=FAILED" in status_text:
```
fengsh27, line 237: Do we need to quit loop if the query failed?

biochatter/api_agent/blast.py
```diff
@@ -0,0 +1,352 @@
...
+                with open(f"{save_path}{file_name}", "w") as file:
+                    file.write("BLAST query FAILED.")
+            elif "Status=UNKNOWN" in status_text:
+                with open(f"{save_path}{file_name}", "w") as file:
+                    file.write("BLAST query expired or does not exist.")
+                raise
```
fengsh27, line 243: is this code intentional?

fengsh27:
Looks good to me except some minor issues

biochatter/api_agent/blast.py
Tatan47: If the query returns a FAILED status the elif is triggered and the loop is ended.

biochatter/api_agent/blast.py
Tatan47: This is a good point. For now i have it like this (i.e. the error is written in the blast result file) because it makes it easier to evaluate the results in downstream testing of the performance of the api agent: **did the query fail or the interpretation of the content of the BLAST file**. Optimally we would indeed handle this differently but it would require another logic. For now i would suggest to keep it as like this.

biochatter/api_agent/blast.py
fengsh27: I'm a little confused, in `elif "Status=FAILED" in status_text:` branch, we only write message to file without including `break` statement. How can we end the loop? Is this intentional?

biochatter/api_agent/blast.py
fengsh27: This also made me confused, why do we need to write result to a file instead of returning it? This makes caller difficult to parse result, particularly for GptConversation, which have to read it and pass it to LLM.
Is there something I might have overlooked?

biochatter/api_agent/blast.py
Tatan47: You are totally correct, I oversaw the missing break! thanks :)

biochatter/api_agent/blast.py
Tatan47: I understand your confusion, the optimal way of handling this would indeed be to implement it as a return instead of writing to file. 
This is a 'legacy code' from my initial implementation where I decoupled the query+write return to file and answer extraction to speed up the testing (GeneTuring test:  https://www.ncbi.nlm.nih.gov/pmc/articles/PMC10054955/). But bottom line: you are right and I will change as soon as i have time for it.

biochatter/api_agent/api_agent.py
```diff
@@ -0,0 +1,151 @@
...
+        question using a BioChatter conversation instance.
+        """
+        try:
+            conversation = self.conversation_factory()
+            return self.query_builder.parameterise_query(question, conversation)
+        except Exception as e:
+            print(f"Error generating query: {e}")
+            return None
+
+    def submit_query(self, api_fields: BaseModel) -> Optional[str]:
```
slobentanzer, line 71: @Tatan47 do we need to rearrange the logic in the APIAgent following the recent changes to returning instead of writing?

slobentanzer:
@Tatan47 I'd be ready to merge into `dev`, let me know your status at your convenience

slobentanzer (approved):
Looks great, many thanks!

PR #193, opened 2024-08-06
fix: resolve an vectorstore issue

fengsh27:


fengsh27:
For some articles, its title length will be more than 1000 characters, which will result in failure in uploading articles

slobentanzer (approved):

PR #197, opened 2024-08-17
feat: introduce use_reflexion to determine whether to use ReflexionAgent to generate Cypher query

fengsh27:


slobentanzer (approved):
Great work, thanks! Bumping to `0.7.3`

PR #198, opened 2024-08-20
fix: Inject kg context even if kg query result is empty

fengsh27:


slobentanzer (approved):
Makes sense, let's try this way. Bumping to `0.7.4`.

PR #200, opened 2024-08-21
refactor(kg): modify kg query response to prompt LLM to include query statement

fengsh27:


slobentanzer (approved):
looks good, bumping patch to `0.7.5`

Issue #224, opened 2024-12-10
Create `anndata` function calling modules


noahbruderer:
This issue is to create a module in the API agent that can generate function calls for [anndata](https://anndata.readthedocs.io/en/stable/api.html) package, starting with combining, reading and filtering.

anisdismail:
Done in https://github.com/biocypher/biochatter/pull/229
- develop a starter for pydantic classes for the anndata reader classes
- develop a tools calling starter code for binding multiple classes
 
Next steps needed: 
- modify the query builder
- modify unit tests for the new structure
- develop a abtract class parser for the pydantic class  

Next sprint: 
- extend it to filtering, mapping and concatenating

PR #229, opened 2024-12-10
Anndata api integration

noahbruderer:
This PR is for the development of a basic anndata tools calling. 
More details are in the issue https://github.com/biocypher/biochatter/issues/224


test/test_integrated_anndata.py
slobentanzer, line 1: What is this script? Does not seem like a unit test, so probably doesn't belong here?

anisdismail:
Added support for tools binding for anndata reading classes + test for parameterize query.

PR #238, opened 2024-12-11
Add the scanpy tool `tl` modules to API agent, next and closing steps. 

bastienchassagnol:


slobentanzer:
could `scanpy` be a dev group dependency (only installed when installing biochatter locally from the repo)? I am not sure we even want it as an optional dependency from pip. Can something be dev and optional at the same time?

test/test_pydantic_class_generation.py
```diff
@@ -0,0 +1,55 @@
+import scanpy as sc
+
+from biochatter.api_agent.generate_pydantic_classes_from_module import generate_pydantic_classes
+
+
+def test_generate_pydantic_classes() -> None:
```
bastienchassagnol, line 6: @slobentanzer is typing relevant for unit functional tests?

slobentanzer: no, this can be deactivated

slobentanzer (changes requested):
Nice work! Todos:
- add docparse library
- move scanpy dependency to dev

bastienchassagnol (commented):
added the docparse-string python module + set aside scanpy into the optional dependencies

slobentanzer (approved):
Great work @bastienchassagnol @mengerj @vd-dragan21 @kvitoslava-yarish :)

Merging

PR #246, opened 2024-12-12
Refactor manual pydantics

mengerj:


slobentanzer (commented):
Looks good from the syntax and logic standpoint, @anisdismail and @noahbruderer should check whether the previous behaviour is retained.

PR #254, opened 2024-12-12
Added create runnable test for Anndata

vd-dragan21:


noahbruderer:
Great work! Thanks :)

PR #259, opened 2025-01-02
feat: support to inject token usage update callback

fengsh27:
### Abstract
This submission introduced support for injecting a token usage update callback into biochatter.

I'm trying to implement a token limitation for the server's community OpenAI key. Its workflow is like the following:

1. biochatter-server will inject token update callback into biochatter. During each chat, biochatter will call the injected callback to update token usage in biochatter-server
2. Biochatter-server will store token usage data in a sqlite3 database
3. If the server community token usage exceeds predefined limit, we will restrict further usage and encourage users to provide their own OpenAI API key for continue access.

This submission also simplified GptConversation constructor by removing redundant arguments ( api_key, is_azure, azure_deployment, azure_endpoint)


slobentanzer (approved):
@fengsh27 thanks, great addition!

Be mindful that I fixed a typo in a function in the `DocumentEmbedder`; `def set_chunk_siue` is now `def set_chunk_size`. If something downstream used the function with typo, this may break. I am bumping minor version for that reason.

I also used the opportunity to fix a lot of Ruff warnings in the two modules. Since the recent change to Ruff, I respond to the warnings that are easily fixed each time I work on a given file.

Will merge and bump to `0.8.0`.

PR #261, opened 2025-01-08
Add the benchmark judgement framework

HansJarchow:
# I have made updates to the following files:
 - `benchmark_utils.py`
 - `conftest.py`
 - `load_dataset.py`
 - `benchmark_longevity_data.yaml`

 *(07.01.24/08.01.24)*

## Updates to `benchmark_utils.py`:

Since responses for the benchmark cases/items are to be generated automatically without requiring detailed analysis of, for example, generated errors, I have separated the creation of a response result file from the creation of other result files (both for RAG and non-RAG responses):

```python
if "responses" in task:
        task_results = return_or_create_response_file(task, model_name)
    elif "rag" in task:
        task_results = return_or_create_rag_response_file(task, model_name)
    else:
        task_results = return_or_create_result_file(task)

        # check if failure group csv already exists
        return_or_create_failure_mode_file(task)

        # check if confidence group csv already exists
        return_or_create_confidence_file(task)
```

Additionally, I implemented the functions `return_or_create_response_file` and `return_or_create_rag_response_file`. Both functions generate the DataFrame in which the model responses are to be stored. They are accompanied by the helper functions `get_response_mode_file_path` and `get_rag_response_mode_file_path`. Any newly created data during a test or run is written to the corresponding files via `write_responses_to_file` and `write_rag_responses_to_file`.

## Updates to `conftest.py`:

To facilitate the collection of model responses during each test run, I introduced a new `fixture`, `multiple_responses`. This fixture collects, formats, and returns the model responses. It functions similarly to `multiple_testing`, but instead of returning a score, it returns a response.

In alignment with the newly created test files (see below), I extended the pytest hook function to include the fixtures used in these tests. These are then parameterized for the new tests.

## Updates to `load_dataset.py`:

I added an additional line to the function `_expand_longevity_test_cases`, which was created during our meeting:

```python
new_case["input"]["prompt"] = new_case["input"]["prompt"].strip()
```

This addition serves to format the synthesized cases (without an appendix; see below).

## Updates to `benchmark_longevity_data.yaml`:

I added all cases, including their rephrasings. Additionally, I introduced the option to include cases without an appendix. This allows for a direct comparison between performance with and without the appendix.

# Newly Created and Updated Files:

- `create_responses.py`
- `create_rag_responses.py`
- `./judgement/`
- `benchmark_longevity_rag_data.yaml`

**Details:**
\
`create_responses.py` generates responses without RAG.
\
`create_rag_responses.py` generates responses with RAG.

Both scripts can be executed using **`pytest`**.

The `./judgement/` directory contains all the scripts and functions required for the “LLM-as-a-Judge” framework, as well as the following subdirectories:

- `./model_eval/`
- `./prompts/`
- `./responses/`
- `Subdirectories:`

`./model_eval/` contains the evaluations generated by the Judge in `.csv` format.

`./responses/ `contains the model responses to be evaluated (results from `create_responses.py` and `create_rag_responses.py`).

`./prompts/` contains the prompts defined for the Judge, stored in `.yaml` format and organized by the specified evaluation criterion. For test purposes the file only includes a prompt for assessing `correctness`.

The scripts are integrated into a **`pytest` framework**. I opted for a separate implementation, distinct from the original **BioChatter-Benchmark workspace**, to leave the existing scripts in that workspace largely unmodified. Perhaps, this separation improves clarity by enabling the Judgement to be executed independently of the “original” testing procedure.

## `conftest.py`

This file includes all Judge models, fixtures, and related configurations.

## `judge_connect.py`

Establishes interaction with the desired model. Currently, only a connection with `OpenAI (GPT)` is defined, but the script is extendable.

## `judgement_utils.py`

Contains all functions for writing result files, reading files, formatting DataFrames, and related operations.

## `test_judgement.py`

Implements the test that executes the Judgement process.

## `benchmark_longevity_rag_data.yaml`

This file contains the cases for evaluating performance under RAG.

# Examples

`./judgement/` contains an example for possible benchmarking results (in `./responses/` and `model_eval`).

I also created `Data_Analysis.ipynb` for first example analyses. The results are not polished and representative, since I conducted the experiment with `257/600` items und `N=1`. **There are some problems with calculating the standard deviation (`std = 0`, for each item). I guess, I would have to calculate the standard deviation before aggregating at the group level.** 

# Final
Overall, I hope that everything has been implemented and documented correctly and in an understandable manner. The Judgement framework should also be modular. This means it could be extended to incorporate additional evaluation criteria or metrics. So far, only a binary format has been integrated.

I guess, I have to take a closer look on the difference between `appendix` and `no_appendix`. **Maybe, it is possible to find more confusing appendices.**


slobentanzer:
Hi @HansJarchow, many thanks for the PR and extensive description! I am still travelling, but on cursory look have some high-level thoughts about the exact prompts that you may already be able to consider. It is mainly related to what you already brought up in coming up with more confusing appendices.

In addition to that, the setup parts that we created quite ad-hoc in out meeting, "plain" and "emotional", I am not certain they reflect the actual extremes we may want to test in the use cases. Did you give it some thought which kinds of setups would be the most extreme? In addition to that, even if we use the plain/emotional distinction as the main two components, it feels that some of the prompts labelled as emotional are not actually phrased that emotionally or colloquially. Before we run the test at a larger scale, it would make sense to be relatively certain that these components (so the `setup` and the `appendix`, mainly, but to a smaller extent also the `background` section and `question`) reflect the spectrum of user queries as extremes, e.g.:
- the very detailed power-user that knows a lot of terminology and also how LLMs work, uses medical vocabulary and so on, versus
- the very naive, colloquial user that asks questions without a deep understanding of how LLMs work and does not know medical terminology.

I think a lot of these decisions come down to experimenting and exploring various prompts to come to some conclusion; it is necessary to do that before we run the full combinatorial benchmark, otherwise we run the risk of having to rerun it. Would you like to think about that and experiment a bit to come up with reasonably "extreme" variations for each of the components? Meanwhile I will look at the code once I'm back.

HansJarchow:
Hi @slobentanzer, Thank you for your response! I want to apologise for the delayed reply; we had quite a few long meetings yesterday.

I’ve also noticed that there isn’t much difference between the `“plain”` and `“emotional”` versions. Similarly, the distinction between `“list”` and `“freetext”` could indeed be more pronounced. **Would it perhaps be possible to use a locally running model (Llama 3.1 8B, Llama 3.2) to adjust the existing versions?** However, I’m unsure whether this might cause the Judge (GPT-4o mini) to rate texts modified by a “non-judge” model differently due to potential bias. But, could it even be noticeable for the judge? An alternative, of course, would be to make the adjustments manually.

I think the difference between `"background"` versions is most likely to impact model performance, as this forms **the core and most extensive part** of a case. It’s also possible that the `appendices` could include “extreme” scenarios, though they **might need to be longer and more complex** to achieve this. The challenge lies in introducing appendices or additional clauses that feel confusing and extreme without adding information that’s directly relevant to the case’s topic.

I’ll try to work on adapting the components.

slobentanzer:
Hi @HansJarchow, agree, but I think one of the points of GSM-Symbolic was that the appendices were not at all very extreme; the point about them was that they would not have confused a human at all, but they somehow confused the LLM. Maybe you can refer back to the GSM-Symbolic paper for the scope and extent of the appendices they used.

For the other components, as before, I think they should try to emulate user behaviours we actually expect; there's no need to come up with something extreme that we wouldn't probably see in the wild. But, for instance, the list-like background could be the well-informed user, while the free text could be more colloquial, not knowing many medical terms and only vaguely describing what the list makes explicit.

slobentanzer:
I had also thought about having an LLM curate the test cases, but I think in this case manual control is better, as you say, we may introduce LLM bias if we do that.

HansJarchow:
Hi @slobentanzer, I’ve put together three examples. I looked into colloquial phrases and generated examples of phrases or sentence beginnings that could be implemented within the topics we discussed, i.e., I used the LLM **to only generate examples for colloquial words or sentence beginnings.**

Before proceeding with the other cases, I wanted to check if this aligns with the direction we’re aiming for:

```python
Example 1: """I’m a 52-year-old guy with a BMI of 27.5. To be honest, 
I’m not a picture of health. Yeah, where to start. I’m pretty much sticked 
to my chair and don’t get much exercise. My diet is not winning any awards – 
it is heavy on processed foods and sugary stuff. It isn’t exactly helping my 
case but I drink alcohol 3-4 times a week. Heart disease and stroke kind 
of run in my family, which concerns me. I got some bloodwork done, and 
honestly, I don’t know what’s going on – my glucose is 105, my LDL 
cholesterol is at 130, HDL is 42 and CRP is 3.2."""

Example 2: """I’m a 34-year-old woman working as a nurse. My BMI is 
21.2. I enjoy regular jogging and yoga workouts and I’m on my feet most of 
the day, so I wouldn’t call myself sedentary. I’ve been following a vegan diet 
for a while now without processed and refined foods, which I thought could 
help to keep things in shape. I’ve never touched a cigarette in my life, but I 
need my daily coffee break. Metabolic diseases run in my family. My mother 
was diagnosed with diabetes type 2 at age 58. My glucose is 90 mg per dl, 
my LDL cholesterol levels show 90 mg/dl.  My CRP is 0.6 and my IGF-1 is 
at 250 µg/l."""

Example 3: """I’m a 68-year-old guy. I’ve got osteoporosis and sarcopenia, so 
my bones aren’t the best these days and my muscle strength just isn’t what it 
used to be. I’m 172 cm tall and currently weigh 59 kg. I’m glued to my chair, so 
I wouldn’t call myself active. I’ve got diabetes and hypertension. My diet isn’t the 
best, no regular meal times and low in protein and fiber. I take Vitamin D and 
Teriparatide, Ramipril and Metformin. But don’t ask me for the dosages. 
Recently, I’ve got some bloodwork done. It shows a cholesterol level of 105 and 
CRP at 3."""
```

HansJarchow:
Hi @slobentanzer, Thank you very much for the feedback! Yes, you're probably right—it would make more sense to implement the separate files into the original framework. I'll give it a try! At the very least, this “exercise” helped me to get (somewhat) comfortable with Pytest. :)

- **Should I create a new branch and open a new PR for these changes?**

I also have a (perhaps slightly…odd) question:
- While experimenting with the `appendices, backgrounds (list, freetext), and setups (plain, emotional)`, I encountered some limitations and insights that seem to be more topic-specific. Should such information be discussed here or via email? I ask because information in public repositories is, well, public. How is this usually handled in projects like this? (I'm still quite “green”/inexperienced in this regard…)

slobentanzer:
> Should I create a new branch and open a new PR for these changes?

The review says "changes requested", meaning that you can apply the requested changes to the current PR; no need to create a new one. Importantly, you need to modify the PR itself, not some branch (the PR technically is a branch).

> Should such information be discussed here or via email?

There are no strict rules; can be either. If you think the content discussion goes too far off-topic, feel free to ask via email or on Zulip first.

HansJarchow:
# Updates
*(13.01.24/14.01.24)*

Hi @slobentanzer, I’ve tried to implement your suggestions, specifically by integrating the separated scripts for judgment into the “original” framework.

## Changes to `conftest.py`:

I’ve defined the judges (currently: `OPENAI_JUDGE`). This includes `METRICS`, which contains the metrics or variables that will later be used to evaluate the model responses. Why did I define `METRICS`? To allow the tests to be parameterized based on the requirements:

```python
OPENAI_JUDGE = [
    "gpt-4o-mini-2024-07-18",
]

JUDGES = OPENAI_JUDGE

METRICS = [
    "correctness",
    "comprehensiveness",
]
```

In addition to a fixture for `METRICS`, I created another `fixture` for the defined `JUDGES`:

```python
# METRICS
@pytest.fixture(params=METRICS)
def judge_metric(request):
    ...

# JUDGES
@pytest.fixture(params=JUDGES)
def judge_name(request):
    ...
```

Another `fixture` was added to call a `conversation object` with the judge. I believe it’s more practical to use the classes already defined in **BioChatter**:

```python
@pytest.fixture()
def judge_conversation(model_name):
    ...
```

Since I created scripts that sequentially generate the responses and evaluations (first generating the `response file` and then the `judgment file`), I had to define a `fixture` that dynamically reads the response file during runtime and passes it to the judgment function:

```python
@pytest.fixture
def test_judge_longevity_responses():
    """Fixture to dynamically load judgment data."""
    path = "./benchmark/LLM_as_a_Judge/responses/"
    if not os.path.exists(path) or not os.listdir(path):
        pytest.skip(f"No files found in directory: {path}")
    
    try:
        JUDGEMENT_DATA = get_judgement_dataset(path)
        return JUDGEMENT_DATA["judgement"]
    except ValueError as e:
        pytest.fail(f"Failed to load judgment data: {e}")
```

I’ve also modified the `pytest_generate_tests` hook function to account for the additional tests.

## Changes to `benchmark_utils.py`:

I initially structured the conditions in `benchmark_already_executed` similarly to the previous pull request. **Why?** My goal is to generate the response and judgment files independently of the `confidence` and `failure mode` files until I find out how considerations associated with those files can also be applied to judgment:

```python
if "judgement" in task:
    task_results = return_or_create_judge_file(task, model_name)
    return False
elif "generation_plain" in task:
    task_results = return_or_create_response_file(task, model_name)
elif "generation_rag" in task:
    task_results = return_or_create_rag_response_file(task, model_name)
else:
    task_results = return_or_create_result_file(task)
```

Additionally, the functions that were temporarily isolated in `judgement_utils.py` have now been integrated into `benchmark_utils.py` under the header `“Functions Judgement”` for easier access.

## "Test Files":
I’ve created scripts to run the response generation and judgment either **sequentially** or **simultaneously**. These scripts include:

- `test_longevity_judge_responses.py`
- `test_longevity_judge_responses_rag.py`
- `test_longevity_judge_responses_simultan.py`
- `test_longevity_judge_responses_rag_simultan.py`

I’m aware the naming isn’t ideal and could be improved, but it should suffice for development purposes.

**Sequential Execution:** \
The scripts `test_longevity_judge_responses.py` and `test_longevity_judge_responses_rag.py` run response generation and judgment *sequentially*. For this, I’ve defined a dependency and an order:

```python
@pytest.mark.dependency(name="generate_responses")
@pytest.mark.order(1)
def test_longevity_response_generation_plain(
    model_name,
    test_create_longevity_responses,
    conversation,
    multiple_responses,
):
    ...

@pytest.mark.dependency(depends=["generate_responses"])
@pytest.mark.order(2)
def test_longevity_judgement(
    judge_name,
    judge_metric,
    test_judge_longevity_responses,
    judge_conversation,
):
    ...
```

All responses are first generated using `test_longevity_response_generation_plain` (distinguished from rag).
Judgment is then executed using `test_longevity_judgement`. Here, the dataset containing the responses is dynamically injected during runtime using the fixture `test_judge_longevity_responses`. This was necessary because `Pytest` parameterizes datasets at the beginning of a test, not during runtime.
The response file is saved in `LLM_as_a_Judge/responses/`. Responses are generated and then evaluated, with the corresponding evaluation file stored in `LLM_as_a_Judge/judgement`.

This flow also applies to `test_longevity_response_generation_rag`.

> At the start of judgment, it’s possible that files from previous runs already exist in the response folder, in addition to the newly created response file. In this case, are all response files reevaluated? No. Only the most recently generated response file is evaluated: 
> ```python
> latest_file = [max([f"{path}/{file}" for file in files], key=os.path.getmtime)]
> ```
> Alternatively, all existing files could theoretically be judged in one run.

**Why an intermediate response file?** It might be useful for additional human evaluation, which shouldn’t be influenced by the judgment. While a response file with judgment could also be edited downstream, having an intermediate file eliminates this extra step.

**Simultaneous Execution:** \
The scripts `test_longevity_judge_responses_simultan.py` and `test_longevity_judge_responses_rag_simultan.py` perform response generation and judgment simultaneously. Responses are generated and immediately evaluated within the same test run. No intermediate response file is created, and the judgment file is directly saved in `LLM_as_a_Judge/judgement/`.

slobentanzer:
Hi @HansJarchow,
this looks great, much better already; well-integrated with the existing framework. I am not sure I understand the sequential testing and reasons behind it. It would have followed the general philosophy of the benchmark to check for the existence of a result (whether it's an intermediary `response` or the final `judgement`) and only execute the particular test that has not been run yet.

```mermaid
flowchart TD
    A[Start with Generated Test Data] --> B[Expand Test Data with\nVariations & Expected Outputs]
    
    B --> C{Check if Responses Already Generated}
    
    C -->|Yes| D{Check if Judgements Already Made}
    C -->|No| E[Generate Multiple Model Responses]
    
    E --> F[Write Responses to File]
    F --> D
    
    D -->|Yes| G[Skip Test]
    D -->|No| H[Run LLM Judge Evaluation]
    
    H --> I[Write Judgements to File]
    
    G --> J[End]
    I --> J
```

The `checks` here can be implemented using the existing process of comparing the task name and a specific hash that was generated for each test case (as a check whether the test case has been modified); `skip_if_already_run`. I would not make it a user task to decide which stages to run, explicitly. This is fairly simple with the response generation (we just need to attach the generated hash to the response file in some way), and the judgment can also be adjusted to follow that paradigm, I believe.

Here's how the integrated test module could look:

```python
"""
TestLongevityResponseAndJudgement
================================

A test class that handles both response generation and judgment evaluation for 
longevity/geriatrics benchmark questions. Can run both stages or just judgment 
stage if responses already exist.

Flow:
1. Setup test data and configurations
2. Check if responses exist
   - If no: generate and store responses
   - If yes: load existing responses
3. Check if judgments exist
   - If no: run judgment evaluation
   - If yes: skip test
"""

class TestLongevityResponseAndJudgement:
    def setup_method():
        """
        Initialize test state:
        - Clear responses list
        - Set default iterations
        - etc.
        """
        
    def test_generate_responses(model_name, test_data, conversation, ...):
        """
        Stage 1: Response Generation
        --------------------------
        - Skip if responses exist for (model, task, hash) combo
        - Generate multiple responses using conversation API
        - Store responses both in memory and to file
        - File format: CSV with metadata (model, task, timestamp, etc.)
        
        Returns: None (responses stored in self.responses and file)
        """
        
    def test_judge_responses(model_name, judge_name, metric, test_data, ...):
        """
        Stage 2: Response Judgment
        ------------------------
        - Skip if judgments exist for (model, task, hash) combo
        - Load responses from file if not in memory
        - For each response:
            - Run multiple judgment iterations
            - Calculate agreement score
        - Store judgments to file
        
        Returns: None (judgments written to file)
        """
        
    # Helper methods
    def _load_responses():
        """Load responses from file if they exist"""
        
    def _format_judgment_prompt():
        """Format prompt for judge based on metric type"""
        
    def _calculate_judgment_scores():
        """Calculate agreement scores across iterations"""

"""
Usage:
------
1. Full pipeline:
   pytest test_longevity_judge_responses_simultan.py

2. Just judgment (with existing responses):
   pytest test_longevity_judge_responses_simultan.py::TestLongevityResponseAndJudgement::test_judge_responses

Required changes to existing codebase:
------------------------------------
1. Modify skip_if_already_run() to handle separate response/judgment checks
2. Add response loading utility
3. Update file paths/naming conventions for better separation
4. Update test configurations for sequential running
"""
```

Is it possible to implement this, or is there some procedural reason to keep the current structure?

HansJarchow:
Hi @slobentanzer, Thank you very much for the detailed blueprint!

I think it shouldn’t be a problem. I’ve already given it some thought and will give it a try. It’s another thing I didn’t know about `Pytest` - organising tests into methods within entire test classes.

If any issues arise, I’ll reach out to you - if that’s okay for you.

HansJarchow:
# Updates
*(15.01.24)*

Hi @slobentanzer, I've tried to integrate the suggestions into the benchmark.

## Benchmark Cases:
Firstly, I worked on making the `freetext` option look more like an actual "free-text." The idea of using this option to represent individuals who are either less specialized or use more colloquial language is a good one. Such individuals might not provide specific units or any values at all. Although I’m still relatively inexperienced, I think it’s unlikely in practical medicine that every patient knows their blood values (or medications). During internships, I often observed that many patients don’t know why they are being treated or what medications they are taking. I also attempted some further rephrasing — for example, turning **“osteoporosis”** into a simple **“bones aren't the best anymore"**.

Secondly, I worked on making the `appendices` a bit more creative. Unfortunately, I’m not yet sure what effects these changes will have on the response behavior of the models. I always worry about possibly becoming “too extreme” or mixing the `appendix` and the actual case too much.

slobentanzer:
As I am going through the cases, noticed some high-level things:
- The system prompts are the same for all cases, right? Should we just define them once and use them in all longevity cases (via some step in preprocessing, like the expansion step)?
- The system prompts don’t have a flattery component (something like “you are a medical expert with specialisation gerontology and longevity and know everything about lifestyle interventions”). While for current top LLMs like 4o this does not make a huge difference, open source models may still be impacted, biasing the benchmark towards the corporate solutions. We could also make the flattery component a part of the benchmark (have system prompts with flattery component and without). Maybe the basic setting could be no system prompts at all, then basic with/without flattery component, and then the “requirements” one (probably only with flattery, as it is the most extensive one). 

I’ll also commit some minor language fixes to the cases as soon as I went through all of them.

slobentanzer:
Some more observations:
- The `individual` field seems to me to refer only to age; rename the field to `age`?
- The case naming is not descriptive; all cases are named `yes or no` and the case ids are integers. Would it not be better to name the cases so it’s obvious at analysis which case we’re looking at (something like `longevity:cushing`)?

Further, are the cases in the RAG tests the same as in the non-RAG tests (as they have the same numbers)? If so, I would suggest to keep them in one YAML, and add the RAG info to the existing cases, and only in the test code turn the usage of RAG data on or off (can be different test methods as well, where the non-RAG test just does not use the RAG info and potentially a different system prompt). Otherwise, the cases may become desynchronised between the two YAML files, and there is no test to make you aware of that.

I have done some minor language correction in the non-RAG cases, but will hold off on the RAG ones until we have clarified that point.

HansJarchow:
Hi @slobentanzer,

Thank you very much for your feedback!

> The system prompts are the same for all cases, right? Should we define them just once and reuse them across all longevity cases (e.g., by including this step in preprocessing, such as during the expansion phase)?

The system prompts are identical across all cases. If it's possible to define them once, we could do so — especially if it contributes to improved clarity and structure.

> The system prompts don’t include a flattery component (e.g., something like, “You are a medical expert specializing in gerontology and longevity and have comprehensive knowledge about lifestyle interventions”).

You’re absolutely right here. Including a flattery component as part of the test is an interesting idea. This would mean the **baseline case** is one without a system prompt and without a flattery component. We could then compare it to:

- A basic version with a flattery component,
- A version with only the system prompt (no flattery), and
- A version with both the system prompt and flattery component.

For the requirements, we would also include both the system prompt and the flattery component. This approach would definitely yield more test items and provide deeper insights into the dependencies on prompt settings, assuming time constraints don’t become an issue.

> The “individual” field appears to refer only to age; should we rename it to “age”?

Absolutely, that would make it much clearer.

> The case naming is not descriptive.
Currently, all cases are named “yes” or “no,” with the case IDs being integers. Wouldn’t it be better to name the cases more descriptively so that it’s immediately clear during analysis which case we’re referring to (e.g., “longevity:cushing”)?

We can definitely make the case descriptions more descriptive. I added the case ids to allow for performance visualization across cases, such as in a heatmap or similar methods.

> Are the cases in the RAG tests the same as those in the non-RAG tests (since the numbers match)?

Yes, they are the same cases and in the same order.

Lastly, I want to apologize for the somewhat confusing structure of the documentation so far (e.g., “individual” instead of “age,” etc.). Until now, I’ve been working on almost everything alone, so I always knew what I meant and where to find things. Thank you so much for the fantastic and close collaboration. It’s a relief not to feel completely helpless anymore 🙂

slobentanzer:
No worries, let me know once you're ready for more input. :)

> For the requirements, we would also include both the system prompt and the flattery component.

I would not create too many different scenarios; it might be that the differences are not that large and so you end up using up a lot of processing for "no difference". I would start with the extremes. No system prompt, basic with and without flattery, and the "full prompt", meaning requirements with flattery. You can still extend later if you feel it would be informative.

HansJarchow:
Alright, I’ll focus on finishing the test class and test methods first. I tend to work on multiple (or too many) things at once, so I need to be careful not to lose track of things 😅

HansJarchow:
# Updates
*(17.01.24)*

Hi @slobentanzer,

I’ve tried to implement the suggestions for the simultaneous execution of response generation and judgements.

To do so, as planned, I also had to make changes to `skip_if_already_run` and `benchmark_already_executed`.

## `benchmark_already_executed`:

I’ve added three additional (optional) arguments. These arguments can be assigned appropriate values if switching from a “normal” benchmark to a judgement. For `mode`, the options are `"response"` (for response generation) and `"judge"` (for the actual judgement):

```python
def benchmark_already_executed(
    model_name: str,
    task: str,
    md5_hash: str,
    mode: Optional[str] = False,
    metric: Optional[str] = False,
    judge_name: Optional[str] = False,
) -> bool:
    """Checks if the benchmark task and subtask test case for the model_name have
    already been executed.

    Args:
    ----
        model_name (str): The model name, e.g. "gpt-3.5-turbo"

        task (str): The benchmark task, e.g. "biocypher_query_generation"

        md5_hash (str): The md5 hash of the test case, e.g.,
            "72434e7a340a3f6dd047b944988491b7". It is created from the
            dictionary representation of the test case.

    Returns:
    -------
        bool: True if the benchmark case for the model_name has already been
            run, False otherwise

    """
    def task_executed(task_results, filters):
        if task_results.empty:
            return False
        return task_results.query(filters).shape[0] > 0

    if mode is False and md5_hash != "?":
        task_results = return_or_create_result_file(task)

        # check if failure group csv already exists
        return_or_create_failure_mode_file(task)

        # check if confidence group csv already exists
        return_or_create_confidence_file(task)

        return task_executed(
            task_results = task_results,
            filters = f"model_name == '{model_name}' and md5_hash == '{md5_hash}'",
        )
    elif mode == "response":
        task_results = return_or_create_response_file(task, model_name)

        return task_executed(
            task_results = task_results,
            filters = f"model_name == '{model_name}' and md5_hash == '{md5_hash}'",
        )
    elif mode == "judge":
        task_results = return_or_create_judge_file(task, model_name)

        return task_executed(
            task_results = task_results,
            filters = f"metric == '{metric}' and evaluated_model == '{model_name}' and model_name == '{judge_name}' and md5_hash == '{md5_hash}'",
        )

    return False
```

If `mode` is left empty (i.e., `False`), the standard benchmark procedure is selected, and the corresponding files are generated. Regarding the test for `md5_hash != '?'`, you probably already know this: at the start of a test, `skip_if_already_run` is executed with default parameters (possibly during the exploration phase of `Pytest` or during test parameterization?). In this case, files for both response/judgement and the standard benchmark are created. However, for better clarity at the data level, I introduced these conditions to separate these cases.

Additionally, I added the function `task_executed`, which handles the filtering of `task_results`. This might make the code slightly clearer regarding the conditions.

For `mode = "judge"`, the data is filtered based on the combination of `model_name` and `md5_hash` as well as `metric` and `evaluated_model`, since this should form a unique and specific combination for every question being evaluated.

## `skip_if_already_run`:

`skip_if_already_run` naturally takes the same arguments as `benchmark_already_executed`. Apart from a minor change to the `“message”`, nothing else has been altered:

```python
if benchmark_already_executed(model_name, task, md5_hash, mode, metric, judge_name):
    message = f"{mode} mode with {metric}" if mode and metric else "Benchmark"
    pytest.skip(
        f"{message} for {task} with hash {md5_hash} with {model_name} already executed",
    )
```

## `test_longevity_judge_responses_simultan.py`:

The structure here should resemble the one you provided. I added a helper method: `_list_files`, which outputs all `.csv` files in a directory ending with `_response.csv`, corresponding to the names of the files created for response generation.

I also had to assign a custom order to the test methods because `Pytest` was unfortunately executing the tests in reverse order (response and then judge). I’m not sure if this would have caused actual problems without dependencies, but for better clarity and transparency in the test process, I preferred to define an order.

**Additionally:** response and judgement files are now saved in `./benchmark/results/`. There’s one response file for each tested model and one judgement file for all evaluated models (this can be switched if necessary).

## Question about benchmark design:

Yesterday, we discussed potential modifications at the level of system prompts. You wrote:

> “Should we just define them once and use them in all longevity cases (via some step in preprocessing, like the expansion step)?”

Do you mean that the system prompts could, for example, be stored externally once and then introduced during preprocessing of the cases, where they could be “expanded”/shuffled?

---
So stupid, I accidentally force pushed 😬...Have I destroyed something now...

slobentanzer:
@HansJarchow force pushing forces your local commit history onto the branch; depending on whether you had my changes there, you may have removed them. I merged again from the location where I had the changes, I think it should be OK. Check the code at HEAD of this branch whether it still reflects your latest additions, but otherwise should be good.

> Do you mean that the system prompts could, for example, be stored externally once and then introduced during preprocessing of the cases, where they could be “expanded”/shuffled?

No, I would still have them in that YAML, I would only define them once (in a separate field at the top; remove them from the cases), and then during data ingestion read that field and combine it with the other parts of each case to yield the same result as now.

```
longevity_geriatric_case_assessment:
  - general_system_messages:
    none: ""
    simple_without_flattery:
      - "You will receive a user question."
      - "The user addresses topics related to aging, longevity medicine, geroscience, and geroprotection."
      - "The user wants to try out interventions relating to one of the mentioned topics."
      - ..
    simple_with_flattery:
      - "You are an expert..."
      - ...
    requirements:
      - "You are an expert..."
      - "The user addresses topics related to aging, longevity medicine, geroscience, and geroprotection."
      - "The user wants to try out interventions relating to one of the mentioned topics."
      - "Please provide a *correct, comprehensive, and useful answer*."
      - "Consider *human interpretability and explainability*."
      - "At the end of your answer, include a final answer using only 'Final answer: yes' if the user is allowed to follow the intervention, or 'Final answer: no' otherwise."
      - "Do NOT add any style or other words."
    explicit:
      - "You will receive a user question."
      - "The user addresses topics related to aging, longevity medicine, geroscience and geroprotection."
      - "The user wants to try out interventions relating to one of the mentioned topics."
      - "Please consider the following requirements which you will have to fulfill in your response:"
      - "*Correctness*: Being correct means being free from error and being as accurate as possible as well as - in scientific terms - to adhere to approved methodological standards in research, that is, in your answer generation."
      - ...

  - case: choice_yes_no
    case_id: "1"
    input:
      prompt:
        setup:
          plain: "Q: My wife told me that I have to lose some weight."
          emotional: "I’ve really let myself go a bit over the past few years. I love food and, well, sometimes I just eat too much. My wife thinks I should lose some weight, and yeah, I think she’s right. If only it were as easy as it always looks on TV. I once read something about caloric restriction—it’s supposed to be pretty effective."
        background:
...
```

HansJarchow:
Hi @slobentanzer, Thank you! From how it looks on GitHub, my last commits are present. I’d like to apologize again. I had been experimenting a bit with an example using commands earlier; and then there was my terminal timeline, and then...the Enter key happened.

> No, I would still have them in that YAML, I would only define them once (in a separate field at the top)...

Ah, now I understand what you mean. Define them once, and during the ingestion process that still occurs during the test (as usual), read in/out the messages defined in the `YAML`.

HansJarchow:
Ok, sorry. I had a look at the current `benchmark-judgement-framework` branch. The last recorded commit for this (`longevity benchmark`) branch is from two weeks ago; I think it was the first one. Or am I mixing everything up now?

PR #262, opened 2025-01-15
feat: protect chat attributes and improve error handling

slobentanzer:
- Add property decorators to protect chat and ca_chat attributes in Conversation ABC
- Make user parameter optional in all conversation classes
- Add clear error messages when chat attributes accessed before initialization
- Reset chat attributes on authentication failure
- Add tests for new chat attribute behavior

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,7 +22,7 @@ jobs:
 
       - name: Install Poetry
         run: |
-          curl -sSL https://install.python-poetry.org | python3 -
+          curl -sSL https://install.python-poetry.org | POETRY_VERSION=1.7.1 python3 -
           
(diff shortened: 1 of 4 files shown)


mengerj (approved):
Looks good to me. I didn't run any tests as I have some other Issues regarding installing the project with Poetry. I see that you changed the version, which was a problem for me before. But now poetry doesn't find my python executable and so far I couldn't fix it.

slobentanzer:
@mengerj I changed the Poetry version in the CI, as they apparently did a major bump and that's not compatible. The project still is based on Poetry >1 <2.

Have you tried [poetry env use](https://python-poetry.org/docs/managing-environments/#switching-between-environments) to tell Poetry where to look?

PR #280, opened 2025-04-25
Add In-Chat Tool Calling Functionalities and refactor llm_connect

fcarli:
### Summary
This pull request introduces two major conceptual enhancements:

- **Module Refactoring:** The existing `llm_connect.py` module is refactored into separate submodules organized by model providers (both for local models and API-based models).
- **In-Chat Tool Calling:** New functionalities for "In-Chat Tool Calling" are, enabling both locally defined tools and MCP compatibility.

### Detailed Changes and Remarks

**In-Chat Tool Usage:**
  - Leverage the `init_chat_model` function from LangChain as a central hub for supporting various models.
  - Define a new `LangChainConversation` class to encapsulate the tool calling logic, ensuring reusability and cleaner code structure. Comprehensive tests for this class (and for generic tool calling functions) are added.

**Local Model Support:**
  - Tool calling also works for local models using Ollama through the `LangChainConversation` (native and non-native) tool invocation.

**Non-Native Model Workaround:**
  - Implement a workaround strategy for models lacking native tool calling support (e.g., Gemma 3).

**Documentation:**
  - Expand documentation with a dedicated section detailing usage instructions and examples for the newly introduced features.

**Backward Compatibility:**
  - Refactoring of `llm_connect.py` is done to preserve compatibility with existing imports, ensuring minimal disruption to users.

**Dependency Updates:**
  - Upgrade LangChain and LangGraph dependencies to accommodate new features. Update `pyproject.toml` and corresponding lock file.
  - Additionally updated Poetry to version 2, acknowledging this introduces a discontinuity. However, reverting to version 1.x remains straightforward if necessary.


slobentanzer:
@fcarli could you resolve the conflicts before I review, please?

slobentanzer (changes requested):
Hi @fcarli, many thanks for the refactor, looking good!

I did only a formal check, everything seems to be in order except the dependency definition in the `pyproject.toml`, don't understand what happened there. I left a specific comment on the file. Once that is fixed, we can merge. :)

As discussed, I did not perform checks on the new logic and tests, as I reserve this for the final merge to `main`.

slobentanzer (approved):
@fcarli thanks; it is weird, I checked the Poetry v2 docs, and they specified the legacy way in the "handling dependencies" section. Now it is different. Anyway, I think the legacy way is much clearer for a project with complex dependencies, so I also prefer this. Hope they don't deprecate it.

Feel free to merge, many thanks! :)

fcarli:
@slobentanzer, before merging into the other branch, I took the opportunity to add the structured outputs functionality. While doing so, I also spotted some bugs in the previous branch and added several tests on the new `LangChainConversation` to be a bit more thorough! Once re-approved, I will merge!

docs/features/structured_outputs.md
```diff
@@ -0,0 +1,437 @@
...
+
+If the LLM natively supports structured outputs (e.g., newer OpenAI, Google models), the `AIMessage` content will typically be a JSON string representation of your Pydantic model. You can then parse this string to get an instance of your model.
+
+## Current Limitations
+
+-   **No Simultaneous Tools and Structured Output**: Currently, you cannot use tools (as described in `tool_chat.md`) and request structured output by passing both `tools` and `structured_model` arguments in the **same** `query()` call. An attempt to do so will raise a `ValueError`.
+
+-   **Sequential Use is Possible**: However, you *can* use tools and structured outputs sequentially. This is a powerful pattern where initial queries can use tools to fetch or compute data, and subsequent queries can process and structure this data using Pydantic models.
+
+    Here's a more detailed biological example:
```
slobentanzer, line 82: Why is this whole section indented by four spaces? Should be a new header and regularly formatted, no?

docs/features/tool_chat.md
```diff
@@ -0,0 +1,387 @@
...
+
+If you wish to inspect the defaults for these prompts, you can print them from
+the conversation object:
+
+```python
+print(convo.general_instructions_tool_interpretation)
+print(convo.additional_instructions_tool_interpretation)
+```
+
+## MCP tools
```
slobentanzer, line 245: should this be higher up?

slobentanzer (commented):
Thanks @fcarli, only two small comments, plus this:

- I changed a couple of docstrings. Typically, mentioning the module, method, etc, is not necessary and just adds to clutter. "This module is responsible for", "This method is to", these kinds of phrasings should be avoided. Consensus is to use imperative: "Provide connectivity to LLMs" or such.
- I had to update the lock file in order for Poetry to let me install the environment. What is your workflow to update the environment when you change dependencies? Ideally, this should be synchronised via the lock file that is included in the repo, so I should not have to run `poetry lock` if I only pull and want to install.

pyproject.toml
```diff
@@ -23,69 +24,76 @@ classifiers = [
     "Natural Language :: English",
     "Topic :: Scientific/Engineering :: Bio-Informatics"
 ]
+dynamic = ["dependencies"]
+dependencies = ["litellm (>=1.68.2,<2.0.0)"]
```
slobentanzer, line 28: there is now a single dependency in this list, should join the others below

docs/features/structured_outputs.md
fcarli: Agreed, reformatted now and made it a little bit nicer!

fcarli:
> Thanks @fcarli, only two small comments, plus this:
> 
> * I changed a couple of docstrings. Typically, mentioning the module, method, etc, is not necessary and just adds to clutter. "This module is responsible for", "This method is to", these kinds of phrasings should be avoided. Consensus is to use imperative: "Provide connectivity to LLMs" or such.
> * I had to update the lock file in order for Poetry to let me install the environment. What is your workflow to update the environment when you change dependencies? Ideally, this should be synchronised via the lock file that is included in the repo, so I should not have to run `poetry lock` if I only pull and want to install.

Agreed on the docstrings, thanks for the clarification!

Regarding the lock file: I believe I was working on something very experimental in a notebook (which I later abandoned), added a few dependencies, and then forgot to sync the lock file. That said, my intention was to ensure everything was properly locked before merging into main. I’ll put more effort in keeping the lock file in sync moving forward to avoid issues like this. Thanks for catching it!

docs/features/tool_chat.md
fcarli: Not sure of what you mean. Do you mean moving all of the MCP sections up?

pyproject.toml
fcarli: Poetry v2 follows PEP 621 and automatically adds new dependencies in the "dependencies" value of the `[project]` section. So, if we want to stick to the old specification (the one using `[tool.poetry.dependencies]`), we need to manually move the packages every time a new one is added.

docs/features/tool_chat.md
slobentanzer: MCP is the de-facto standard for tool binding and has a lot of momentum; if I read a tool calling doc, I expect to find the most relevant / current functions at the top. Most users will probably want to use this one, so it makes sense to have it higher up, IMO.

pyproject.toml
slobentanzer: No, that's not completely accurate.

You can add a dependency to any `pyproject.toml` section and run `poetry lock` to update the lock file with this suggested addition (this is alternative to `poetry add`). I would not suggest moving them manually without updating the lock file with `poetry lock`.

I don't know whether there is any support for managing dependencies (including optional ones) in the new approach with PEP 621, need to look into that. Currently it looks like a step backwards to me, but maybe they have thought of it in some way.

docs/features/tool_chat.md
fcarli: Moved up!

pyproject.toml
slobentanzer: Yeah I got that, no worries. I just meant I would proceed by adding the library into the `pyproject.toml` in the correct section, saving the file, and then running `poetry lock`. I would not use the add command in this case.

But yes, I still need to check whether it would be better to go to the PEP 621 way; I just haven't had time to go in depth on what it would mean for managing optional dependencies effectively.

slobentanzer (approved):
Looks good now! Feel free to merge :)

PR #290, opened 2025-04-29
Create the final benchmarking framework for longevity experiments

HansJarchow:
**I hope, it works now.**

*(29.04.25)*

**This push/pull contains all benchmarking data.**


slobentanzer (commented):
Hi @HansJarchow, looks good!

Just one question: you have added columns to the result files (like the distractor and type) in the benchmark utils. Does that affect the other benchmarks? What happens if there are results from another benchmark already, but the CSVs don't have these columns, did you check?

HansJarchow:
Hi @slobentanzer,

Thanks for your response! Are you referring to backward compatibility, or cases where someone might use an external CSV that doesn't include these columns? Maybe I could add any missing columns with default values during benchmarking - if there aren't existing?

slobentanzer:
Hi @HansJarchow, feel free to merge into `longevity-benchmark` if compatibility with previous benchmark categories is still given. We can go over it one more time when we do the PR to main from there.

PR #292, opened 2025-05-06
fix: anthropic does not allow system message to appear in the middle of messages list

fengsh27:


slobentanzer (approved):
@fengsh27 thanks, that was quick!

PR #294, opened 2025-05-09
Adapt docs structure

slobentanzer:
This PR aligns the documentation layout with the ecosystem layout to prepare for the merging of documentations into a unified landing page. Sections are renamed and reorganised to reflect the same high-level concept in both biocypher and biochatter.


slobentanzer:
@fcarli FYI, since you updated docs in your work, too We need to carefully check the rearrangement so no new content gets lost from both sides when we PR for `0.10`.

PR #303, opened 2025-06-03
Minor update to conform BioChatter with new BioChatter-light refactor

fcarli:
## Summary of Changes

This PR introduces support for configurable model providers in the BioChatter framework, moving away from the hardcoded OpenAI dependency to a more flexible system that supports multiple LLM providers.

(prepares BioChatter update introduced in https://github.com/biocypher/biochatter-light/pull/63)

### Key Changes:

**1. Model Provider Parameter Addition**
- Added `model_provider` parameter to core classes: `DatabaseAgent`, `BioCypherPromptEngine`, and `RagAgent`
- This allows users to specify which LLM provider to use (e.g., "openai", "google_genai", "anthropic")

**2. Default Model Change**
- Changed default model from OpenAI's `gpt-3.5-turbo` to Google's `gemini-2.0-flash` 
- Updated default model provider to `google_genai`

**3. LangChain Integration**
- Replaced direct `GptConversation` usage with `LangChainConversation` in `BioCypherPromptEngine._get_conversation()`
- This enables support for multiple providers through LangChain's unified interface

**4. API Updates**
- **DatabaseAgent**: Now requires `model_provider` parameter in constructor
- **BioCypherPromptEngine**: Added `model_provider` parameter with default "google_genai"
- **RagAgent**: Added `model_provider` parameter with default "google_genai"

**5. Backward Compatibility Considerations**
- The changes are breaking for existing code that doesn't specify the model provider
- Users will need to update their code to include the `model_provider` parameter when creating `DatabaseAgent` instances

**6. Test Updates**
- Updated all test files to include the new `model_provider` parameter
- Added specific tests to verify model provider functionality
- Tests cover both the new default (Gemini) and custom provider scenarios

### Migration Impact:
Existing users will need to update their code when creating `DatabaseAgent` instances:

```python
# Before
agent = DatabaseAgent(
    model_name="gpt-4",
    connection_args=conn_args,
    # ... other params
)

# After  
agent = DatabaseAgent(
    model_provider="openai",  # New required parameter
    model_name="gpt-4",
    connection_args=conn_args,
    # ... other params
)
```

This change significantly improves the flexibility of the BioChatter framework by allowing users to choose from various LLM providers while maintaining a consistent API interface.


slobentanzer:
@fcarli thanks, great addition. Separating vendor and model is more transparent. Did you check that the benchmark still works? I don't see changes in the benchmark code but imagine this might be relevant there in the model selection.

Since this is a change in user-facing API, I suppose we should bump minor version to `0.11.0`?

fcarli:
> @fcarli thanks, great addition. Separating vendor and model is more transparent. Did you check that the benchmark still works? I don't see changes in the benchmark code but imagine this might be relevant there in the model selection.
> 
> Since this is a change in user-facing API, I suppose we should bump minor version to `0.11.0`?

I checked the benchmark, and I would say that everything seems to be working correctly (at least with the current pyconf.py). I ironed out some additional inconsistencies with the `GptConversation` and the `AnthropicConversation`. The only issue I'm encountering is with the `test_text_image_multimodality.py` benchmark, for which I'm missing the `benchmark/data/source_data_clip` directory.

slobentanzer:
You mean `conftest.py`?

fcarli:
> You mean `conftest.py`?

Yeah sorry

slobentanzer:
So did you not have to change anything in the benchmark? I am not sure now whether this is actually a breaking change or not. I would have thought the additional model provider in the params would necessitate an update in the initialisation of a model.

fcarli:
> an update in the initialisation of a model.

The additional model provider is added only to the `DatabaseAgent` class (modified for alignment with the `BioChatter-light` refactor), and I would say this does not influence the benchmark.

fcarli:
Agreed!

PR #313, opened 2025-06-18
feat: support multiple images in conversations and add PDF-to-image utility

fcarli:
## Summary

This PR enhances BioChatter's multimodal capabilities by adding support for **multiple images** in conversations and introducing **PDF-to-image conversion** functionality.

### Key Changes:

1. **Multiple Image Support**: 
   - Updated `append_image_message()` and `query()` methods to accept either a single image URL or a list of image URLs
   - Maintains backward compatibility with existing single-image functionality
   - Added `local` parameter to the `query()` method for proper handling of local vs remote images

2. **PDF-to-Image Conversion**:
   - Added new `pdf_pages_to_images()` utility function in `biochatter/utils.py`
   - Converts PDF documents into individual page images using PyMuPDF and PIL
   - Supports configurable DPI, image format, and output naming
   - Enables document question-answering by passing PDF pages as images to multimodal models

3. **Documentation & Testing**:
   - Updated multimodal documentation with examples for both features
   - Added comprehensive test coverage for single/multiple image handling
   - Tests cover local files, remote URLs, and mixed scenarios

### Benefits:
- Enables analysis of multiple images simultaneously (e.g., comparing figures, processing document pages)
- Provides fast document Q&A by leveraging multimodal models' image processing capabilities
- Offers an alternative to traditional PDF parsing for noisy documents with embedded images

The changes are fully backward compatible and extend BioChatter's multimodal functionality for more complex use cases involving multiple images and PDF documents.


slobentanzer:
@fcarli looks good; only one comment on the param name. As it seems a minor update, do you suppose we bump patch?

slobentanzer:
@fcarli bump

VladSam42:
@fcarli, the architecture can be improved a bit because the current approach with the `local` flag can break integration with frontends later. A better approach would be:
- enforce accepting only `image_urls: str | list[str]` (either web urls or base64 encoded0, and also raise errors if incorrect. The whole concept of "local files" shouldn't exist at this level and providing the correct url should be handled by the UI layer.
- defer conversion to modules outside; the same way you also kept pdf conversion outside

biochatter/llm_connect/conversation.py
```diff
@@ -501,10 +501,10 @@ def append_user_message(self, message: str) -> None:
     def append_image_message(
         self,
         message: str,
-        image_url: str,
+        image_url: str | list[str],
```
VladSam42, line 504: enforce accepting only image_urls: str | list[str] (either web urls or base64 encoded0, and also raise errors if incorrect. The whole concept of "local files" shouldn't exist at this level and providing the correct url should be handled by the UI layer.

VladSam42 (changes requested):

PR #316, opened 2025-07-07
Improve prompt structure and reduce hallucination risk in web API agents

peymanvahidi:
In this PR we address some improvements to the prompt templates in the web API agents (BLAST, OncoKB and bio.tools) that should help to improve the response quality of the agent.

Summary of the changes:

- Moved context before questions in all summary prompts, this way we reduce the risk of prioritizing the question and skimming the context based on that. Also, model gets all the background info first, then receives the question which follows the best practices in writing prompt templates.
- Moved all the system prompts to the ChatPromptTemplate instead of mixing them in the user prompts.
- Replaced "world class", "knows everything" language with "experienced" and "specializing in" or in query prompts used "system designed to" instead of "world class algorithm".

These changes should reduce hallucination risks and encourage the model to more evidence-based responses.

### DIFF
--- a/biochatter/api_agent/web/bio_tools.py
+++ b/biochatter/api_agent/web/bio_tools.py
@@ -20,7 +20,7 @@ from biochatter.api_agent.base.agent_abc import (
 )
 
 BIOTOOLS_QUERY_PROMPT = """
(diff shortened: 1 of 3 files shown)


slobentanzer:
@fcarli would you like to review this?

PR #304, opened 2025-06-06
Introduce OpenRouterConversation with Force Tool Call and Token Counting Overhaul

fcarli:
## Summary of Changes

This PR introduces significant enhancements with three key features:

### 1. Introduction of the OpenRouterConversation Class

The `OpenRouterConversation` class was introduced by extending the existing `LangChainConversation` framework (and therefore automatically inheriting Tool Usage and Structured Outputs). The implementation includes:

- **New OpenRouter-specific Chat Client**: Created `ChatOpenRouter` class that extends LangChain's `ChatOpenAI` with OpenRouter-specific configuration (API endpoint: `https://openrouter.ai/api/v1`)
- **Seamless Integration**: The `OpenRouterConversation` class inherits from `LangChainConversation`, maintaining consistency with the existing conversation architecture while providing OpenRouter-specific functionality
- **API Key Management**: Implements robust API key handling with environment variable support (`OPENROUTER_API_KEY`) and proper error handling for authentication failures
- **Comprehensive Testing**: Added extensive test coverage including message formatting, API key validation, tool calling, and error scenarios

### 2. Force Tool Call Bypassing Controls

A new **`force_tool`** parameter was added across conversation classes to enable bypassing normal tool call controls. The implementation preserves the existing tool call framework while providing an override mechanism when needed.

### 3. Standardized Token Counting Implementation

Token counting was completely overhauled to provide consistent behavior across all conversation types:

- **Universal Token Extraction**: Implemented `_extract_total_tokens()`, `_extract_input_tokens()`, and `_extract_output_tokens()` methods that handle diverse token usage formats from different providers (OpenAI, Anthropic, Gemini, Ollama, LiteLLM, etc.)
- **Usage Metadata Integration**: Leveraged LangChain's `usage_metadata` attribute on `AIMessage` objects to extract token counts in a standardized way
- **Provider-Agnostic Design**: The system now seamlessly handles various token format patterns:
  - OpenAI/Azure: `{"prompt_tokens": X, "completion_tokens": Y, "total_tokens": Z}`
  - Anthropic: `{"input_tokens": X, "output_tokens": Y}` (calculates total)
  - Gemini: `{"total_tokens": Z}` (direct extraction)
  - Ollama: Integer values (direct usage)
  - LiteLLM: `{"input_tokens": X, "output_tokens": Y, "total_tokens": Z}`
- **Integer Return Format**: All conversation classes now return token usage as integers rather than complex dictionaries, providing a cleaner and more consistent API
[...]

PR #317, opened 2025-07-10
fix: remove return of token usage for the correcting agent for BioChatter light

fcarli:
Erroneously, https://github.com/biocypher/biochatter/pull/304 introduced a return of the token usage for the correcting agent, breaking compatibility with the BioChatter-light interface. This PR addresses the issue by reverting to the original behavior.

### DIFF
--- a/biochatter/llm_connect/langchain.py
+++ b/biochatter/llm_connect/langchain.py
@@ -270,7 +270,7 @@ class LangChainConversation(Conversation):
         response = self.ca_chat.invoke(ca_messages)
 
         correction = response.content
-        token_usage_raw = response.usage_metadata
-        token_usage = self._extract_total_tokens(token_usage_raw)
+        # token_usage_raw = response.usage_metadata
+        # token_usage = self._extract_total_tokens(token_usage_raw)
 
-        return correction, token_usage
+        return correction


slobentanzer:
@fcarli why erroneously? I thought you wanted to change / simplify that 😅

The correction functionality is outdated anyways, we should remove that soon and not spend too much time on maintaining it (IMO). Should be much more generic in an agentic setting. Happy to either roll back or change it in biochatter light. You can make that decision.

Minor comment: make sure to sync with the current version of `main` before branching, this PR is outdated.

fcarli:
> @fcarli why erroneously? I thought you wanted to change / simplify that 😅
> 
> The correction functionality is outdated anyways, we should remove that soon and not spend too much time on maintaining it (IMO). Should be much more generic in an agentic setting. Happy to either roll back or change it in biochatter light. You can make that decision.
> 
> Minor comment: make sure to sync with the current version of `main` before branching, this PR is outdated.

The previous PR simplified token counting across providers for the primary query (that is still there). I had also added token counting for the correcting agent, which wasn’t previously tracked. I’m now reverting just that part, as it’s the lowest-effort change and I agree with you that the current functionality is outdated.

I think that we should, at some point, address the return issue as I introduced in https://github.com/biocypher/biochatter/pull/297 because it will probably make the code more robust to these types of issues.

PR #324, opened 2025-12-02
Update selector_agent.py: prevent LangChainDeprecationWarning from being issued

hdraisma:
Prevent the following "LangChainDeprecationWarning" from being issued:
```
[~]/biochatter/biochatter/llm_connect/conversation.py:34: LangChainDeprecationWarning: As of langchain-core 0.3.0, LangChain uses pydantic v2 internally. The langchain_core.pydantic_v1 module was a compatibility shim for pydantic v1, and should no longer be used. Please update the code to import from Pydantic directly.

For example, replace imports like: `from langchain_core.pydantic_v1 import BaseModel` with: `from pydantic import BaseModel`
or the v1 compatibility namespace if you are working in a code base that has not been fully upgraded to pydantic 2 yet.         from pydantic.v1 import BaseModel

  from biochatter.selector_agent import RagAgentSelector
```


VladSam42:
fixed it in another PR

PR #332, opened 2026-06-08
Fix: strip markdown code fences from Cypher query in DatabaseAgent

rupshali21999:
Problem: BioCypherPromptEngine.generate_query() sometimes returns Cypher wrapped in markdown code fences (```cypher). When DatabaseAgent._generate_query() passes this directly to driver.query(), Neo4j raises a CypherSyntaxError because it cannot parse the backticks.

Fix: Strip markdown code fences and any non-Cypher preamble from the generated query before passing it to Neo4j.

### DIFF
--- a/biochatter/database_agent.py
+++ b/biochatter/database_agent.py
@@ -78,12 +78,9 @@ class DatabaseAgent:
             query = self.prompt_engine.generate_query(query)
             query = query.strip()
             if "```" in query:
-                m = re.search(r"```(?:neo4j|cypher)?\s*(.*?)```", query, re.DOTALL)
+                m = re.search(r"```(?:\w+)?\s*(.*?)```", query, re.DOTALL)
                 if m:
                     query = m.group(1).strip()
-            cs = re.search(r"(MATCH|CALL|WITH|CREATE|MERGE)\s", query, re.IGNORECASE)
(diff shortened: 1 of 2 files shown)


slobentanzer:
@rupshali21999 thanks a lot for the PR! When adding new functionality, it is imperative to test it if the behaviour that should be fixed is not already tested for. This is highly likely, because otherwise some tests would have already failed in the current CI suite.

Could you please add test coverage for this string sanitisation behaviour? One way to think about it is to add tests that will fail in the state of the code before your introduced change. Some people work by adding these tests first, and then the implementation that fixes them (test-driven development). I do like that approach as well and try to do it as often as possible.

Further, can you think of a generalisation of the tests and implementation, to be applicable beyond the Cypher scope; for instance, any query language? The prompt generation accepts other languages (e.g., Postgres / SQL dialects).

rupshali21999:
@slobentanzer I've updated the PR with:

Generalised the fix to work for any query language (removed Cypher-specific keyword detection)
Added 4 tests covering: markdown fences with Cypher, markdown fences with SQL, no language tag, and clean query passing through unchanged

slobentanzer:
@rupshali21999 please change the target of this PR to `staging` as well (same as #334)

